PR #7, opened 2022-10-03
Fix typos and duplicate entry in results

fabianvf:


engine/engine.go
```diff
@@ -91,12 +91,10 @@ func (r *ruleEngine) RunRules(ctx context.Context, rules []Rule) {
 			case response := <-ret:
 				if !response.conditionResponse.Passed {
 					responses = append(responses, response)
-
 				} else {
 					// Log that rule did not pass
 					fmt.Printf("Rule did not error")
 				}
-				responses = append(responses, response)
```
fabianvf, line 99: I assume we don't actually want to do this

fabianvf: ie we should only return violations and not every result

engine/conditions.go
```diff
@@ -2,15 +2,15 @@ package engine
 
 import "fmt"
 
-type CondtionResponse struct {
+type ConditionResponse struct {
```
shawn-hurley, line 5: Never fails

engine/engine.go
shawn-hurley: I think so, but I also think that at like a debug level low enough I could see us logging these

shawn-hurley (approved):

PR #8, opened 2022-10-03
adding logger and piping into rpc client

### LINKED ISSUE #2: We need to choose a logger, and determine default logging behavior
A logger that can be enabled to run with user-readable logs and structured logging as well as levels to aid in debugging I think are the primary requirements

shawn-hurley:
Signed-off-by: Shawn Hurley <smhurley00@gmail.com>

partial-fix #2


fabianvf (approved):
/lgtm

PR #14, opened 2022-10-07
Adding the java provider implementation

### LINKED ISSUE #10: Actually Implement the Java References Evaluation

shawn-hurley:
added ability to cancel commands, but I think this needs to be more robust before marking as done.

Signed-off-by: Shawn Hurley <smhurley00@gmail.com>

fixes #10


provider_settings.json
```diff
@@ -1,12 +1,12 @@
 [
```
pranavgaikwad, line 1: @shawn-hurley can we make this file a sample and add the actual file to gitignore so you don't have to check in your local paths in repo?

pranavgaikwad (changes requested):

provider_settings.json
shawn-hurley: Fixed in the other PR.

pranavgaikwad (approved):

PR #19, opened 2022-10-20
Add pipelining between conditions

fabianvf:

PR #20, opened 2022-10-24
Pipelining

shawn-hurley:
Following on to #19 and carrying patches. 

The TODO's outstanding after this:

1. Decide on templating language and/or implementation. Using go templates is probably not the best thing to do.
2. Add the ability for other providers to use pipelines (java/golang etc).
3. Find the correct way for a provider/capability to document its template context.

### DIFF
--- a/engine/conditions.go
+++ b/engine/conditions.go
@@ -43,13 +43,25 @@ func (a AndCondition) Evaluate(log logr.Logger, ctx map[string]interface{}) (Con
 		return ConditionResponse{}, fmt.Errorf("conditions must not be empty while evaluating")
 	}
 
-	fullResponse := ConditionResponse{Passed: true}
+	fullResponse := ConditionResponse{
+		Passed:              true,
+		ConditionHitContext: []map[string]string{},
+		TemplateContext:     map[string]interface{}{},
+	}
 	for _, c := range a.Conditions {
 		response, err := c.ProviderSpecificConfig.Evaluate(log, ctx)
+		if err != nil {
+			return ConditionResponse{}, err
+		}
 
-		// Short cirtcut loop if one and condition fails
 		if !response.Passed {
-			return response, err
+			fullResponse.Passed = false
+		}
+
+		copy(fullResponse.ConditionHitContext, response.ConditionHitContext)
+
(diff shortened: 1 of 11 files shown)


engine/conditions.go
```diff
@@ -23,48 +34,113 @@ type Rule struct {
 }
 
 type AndCondition struct {
-	Conditions []Conditional
```
fabianvf, line 26: Do we want to add context gathering to the conditions inside and/or?

shawn-hurley: I think you are saying that inside of the or/and that we will add `As` blocks.

My initial reaction is no because it will make things more complicated and make it easier to overshadow contexts. 

If you want to add to the context, you must use a chain. To be specific, I will add a test that checks for this.

shawn-hurley: and example is:


```yaml
- message: POM XML dependencies
  when:
    chain:
    - or:
      - chain:
        - builtin.file: pom.xml
          as: poms
        - builtin.filecontent: TODO(\(.*\))?:?(.*)?
      - builtin.xml:
          xpath: "//dependencies/dependency"
          filepaths: '{{ index .poms "filepaths" }}'
        from: poms
```

shawn-hurley: this is the golang output:

```
 when:
      chain:
      - from: ""
        as: ""
        providerspecificconfig:
          or:
          - from: ""
            as: ""
            providerspecificconfig:
              chain:
              - from: ""
                as: poms
                providerspecificconfig:
                  client: {}
                  capability: file
                  conditioninfo: pom.xml
              - from: ""
                as: ""
                providerspecificconfig:
                  client: {}
                  capability: filecontent
                  conditioninfo: TODO(\(.*\))?:?(.*)?
          - from: poms
            as: ""
            providerspecificconfig:
              client: {}
              capability: xml
              conditioninfo:
                filepaths: '{{ index .poms "filepaths" }}'
                xpath: //dependencies/dependency
```

pranavgaikwad (approved):
VISACK, no objections to merging if this is blocking further progress. can come back and address issues if any pop up

fabianvf:
Yep I meant to lgtm this but retroactive :+1: on merging

Issue #24, opened 2022-10-25
Find the correct way for a provider/capability to document its template context.


shawn-hurley:
Today, a provider is responsible for adding extra information for its particular capability. We need an easy way for these things to provide this information, such that someone can write rules against this information.

pranavgaikwad:
Fixed via #56

PR #39, opened 2022-11-15
Introduce capability struct with provider specific template context schema

pranavgaikwad:
Addresses https://github.com/konveyor/analyzer-lsp/issues/24

~openapi-go does not support v3, I used it as it seemed sufficient for our use-case. Open to discussing other options~

Updated to use kin-openapi/openapi3


fabianvf (commented):
looks good to me, do we want to add a way to print the templatecontext schemas as JSON for users to introspect here or do that in a follow-up?

pranavgaikwad:
@fabianvf openapi3.Schema implements MarshalJSON(), do we want to explicitely add something?

fabianvf:
@pranavgaikwad yeah we don't need to reimplement that, I mean more from the CLI side something to display this output

provider/provider.go
```diff
@@ -19,6 +20,8 @@ import (
 type Client interface {
 	Capabilities() ([]string, error)
 
+	TemplateContext() (openapi3.Schema, error)
```
shawn-hurley, line 23: I wonder if this should be a part of the Capabilities return and if we just make that struct. I could also see it being helpful to have a context per capability.

fabianvf: I like that idea

pranavgaikwad:
@fabianvf I will address the CLI thing in a follow-up. I think the CLI deserves its own separate issue.

provider/lib/lib.go
```diff
@@ -13,6 +14,11 @@ var builtinConfig = Config{
 	Location: ".",
 }
 
+type Capability struct {
+	Capability      string
```
shawn-hurley, line 18: Capability.Capability stutters I think, I think that we maybe want it to be the Name of the cap?

shawn-hurley (approved):
One naming concern but other than that looks good to me.

PR #40, opened 2022-11-17
If a single matching file fails, continue looping over remaining files rather than erroring

fabianvf:

### DIFF
--- a/provider/builtin/provider.go
+++ b/provider/builtin/provider.go
@@ -142,9 +142,14 @@ func (p *builtinProvider) Evaluate(cap string, conditionInfo []byte) (lib.Provid
 		for _, file := range xmlFiles {
 			f, err := os.Open(file)
 			doc, err := xmlquery.Parse(f)
+			if err != nil {
+				// TODO(fabianvf) - do we want to error, log or continue here?
+				continue
+			}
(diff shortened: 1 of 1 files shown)


provider/builtin/provider.go
```diff
@@ -142,9 +142,14 @@ func (p *builtinProvider) Evaluate(cap string, conditionInfo []byte) (lib.Provid
...
 			doc, err := xmlquery.Parse(f)
+			if err != nil {
+				// TODO(fabianvf) - do we want to error, log or continue here?
+				continue
+			}
 			list, err := xmlquery.QueryAll(doc, query)
 			if err != nil {
-				return response, err
+				// TODO(fabianvf) - do we want to error, log or continue here?
+				continue
```
shawn-hurley, line 152: Can we log the error as well? probably pretty high verbosity but will help IMO

shawn-hurley: Can we add an example of this failing to demo A for integration testing?

fabianvf: I can add a log. I can add a test too though it was failing for me on the `rule-example.yaml` already, did you see that behavior as well or no?

fabianvf: I have no idea why this is failing for me tbh

PR #49, opened 2022-11-28
Passed -> Matched, for clarity

fabianvf:

### DIFF
--- a/demo-output.yaml
+++ b/demo-output.yaml
@@ -27,11 +27,6 @@
       innerText: "\n      io.fabric8\n      kubernetes-client-api\n      6.0.0\n    "
       matchingXML: <groupId>io.fabric8</groupId><artifactId>kubernetes-client-api</artifactId><version>6.0.0</version>
   extras: []
-- ruleID: chain-pom-002
-  description: ""
-  category: ""
-  incidents: []
-  extras: []
 - ruleID: file-001
   description: ""
   category: ""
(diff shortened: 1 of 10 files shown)


fabianvf:
I think tests may be failing at least in part because the tests are wrong, going to double check

demo-output.yaml
```diff
@@ -149,6 +144,35 @@
       file: file:///analyzer-lsp/examples/golang/main.go
       lineNumber: 9
   extras: []
+- ruleID: lang-ref-001
```
fabianvf, line 147: I think this should have been there all along

demo-output.yaml
```diff
@@ -27,11 +27,6 @@
       innerText: "\n      io.fabric8\n      kubernetes-client-api\n      6.0.0\n    "
       matchingXML: <groupId>io.fabric8</groupId><artifactId>kubernetes-client-api</artifactId><version>6.0.0</version>
   extras: []
-- ruleID: chain-pom-002
```
fabianvf, line 30: I don't think anything I did should lead to this being excluded?

shawn-hurley: with no incidents, I wonder if we are returning the wrong value in the chain?

shawn-hurley:
This looks good to me thank you this is going to make everything much clear

demo-output.yaml
shawn-hurley: or if it really should not be matched

fabianvf:
@shawn-hurley is that approval to merge?

shawn-hurley:
I think so yeah

provider/builtin/provider.go
```diff
@@ -151,14 +151,17 @@ func (p *builtinProvider) Evaluate(cap string, conditionInfo []byte) (lib.Provid
...
 		query := cond.XML.XPath
+		fmt.Printf("\n\n\n Search Pattern: %v", cond.XML)
 		if query == "" {
 			return response, fmt.Errorf("Could not parse provided xpath query as string: %v", conditionInfo)
 		}
 		//TODO(fabianvf): how should we scope the files searched here?
 		var xmlFiles []string
 		if len(cond.XML.Filepaths) == 0 {
 			pattern := "*.xml"
+			fmt.Printf("\n\n\n Search Pattern: %v", xmlFiles)
```
shawn-hurley, line 162: 👍🏻 way to remove debugging!

PR #57, opened 2022-12-21
:sparkles: Add OpenTelemetery tracing

### LINKED ISSUE #35: Adding Tracing in engine and providers
One of the concerns could be the performance of a given provider. We will want the ability to quickly diagnose if the time taken is in the engine or a given provider. 

We will also want to see the traces from our providers to see if anything is taking longer than it should.

pranavgaikwad:
Fixes #35 

jaeger


pranavgaikwad:
I think @shawn-hurley needs to re-build the base jdtls server image. go-toolset in the current version probably has go 1.17

PR #61, opened 2023-01-18
Produce hierarchical dependency output

### LINKED ISSUE #59: Output dependencies as a linked list

pranavgaikwad:
fixes #59 

- Changes dependency provider interface to return a map of list of dependencies where each key is the direct dependency of the application and indirect dependencies form the list of values for that key. Note that transitive / indirect dependencies are flattened into a list.
- Updates golang / java dep providers to parse dependency in the expected format

In future, the same map can be used to return an adjacency list instead of an absolute list to further maintain the hierarchy within transitive / indirect deps.

### DIFF
--- a/dependency/main.go
+++ b/dependency/main.go
@@ -1,16 +1,33 @@
 package main
 
 import (
+	"flag"
 	"fmt"
+	"os"
 
 	"github.com/konveyor/analyzer-lsp/dependency/provider/java"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	outputFile = flag.String("output", "output.yaml", "path to output file")
 )
 
 func main() {
+	flag.Parse()
+
 	p := java.GetDepProvider()
-	deps, err := p.GetDependencies("/home/shurley/repos/analyzer-lsp/examples/java/pom.xml")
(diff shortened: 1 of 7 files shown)


dependency/main.go
```diff
@@ -1,16 +1,33 @@
...
+var (
+	outputFile = flag.String("output", "output.yaml", "path to output file")
 )
 
 func main() {
+	flag.Parse()
+
 	p := java.GetDepProvider()
-	deps, err := p.GetDependencies("/home/shurley/repos/analyzer-lsp/examples/java/pom.xml")
+	deps, err := p.GetDependencies("/home/pranav/Projects/analyzer-lsp/examples/java/pom.xml")
```
shawn-hurley, line 20: we maybe want to make this a relative path?

shawn-hurley:
/lgtm with a quick pass and I didn't run it. 


The only large comment is a test for golang as well? but I didn't create one either so I get it.

pranavgaikwad:
@shawn-hurley added tests for go provider, updated to use relative path in main.go.

shawn-hurley (approved):

PR #68, opened 2023-01-31
Implement dependency conditions

### LINKED ISSUE #44: Handle dependency analysis
Analyzers need an ability to perform dependency analysis.

fabianvf:
Closes #44 


engine/engine.go
```diff
@@ -8,11 +8,13 @@ import (
 
 	"github.com/cbroglie/mustache"
 	"github.com/go-logr/logr"
+	"github.com/konveyor/analyzer-lsp/dependency/dependency"
 	"github.com/konveyor/analyzer-lsp/hubapi"
 )
 
 type RuleEngine interface {
 	RunRules(context context.Context, rules []Rule) []hubapi.Violation
+	AddDependencies(provider string, deps map[dependency.Dep][]dependency.Dep)
```
shawn-hurley, line 17: I think remove this?

provider/provider.go
```diff
@@ -133,3 +139,62 @@ func GetProviderClient(config lib.Config) (Client, error) {
...
+		constraintPieces = append(constraintPieces, "< "+dc.Upperbound)
+	}
+	constraints, err := version.NewConstraint(strings.Join(constraintPieces, ", "))
+	if err != nil {
+		return resp, err
+	}
+
+	resp.Matched = constraints.Check(depVersion)
+	resp.Incidents = []engine.IncidentContext{engine.IncidentContext{
+		FileURI: matchedDep.Location,
```
fabianvf, line 188: I think location means something other than what I thought it did

rule-example.yaml
```diff
@@ -73,3 +73,27 @@
...
+          name: golang.org/x/text
+          upperbound: v0.3.8
+      - go.dependency:
+          name: k8s.io/apimachinery
+          lowerbound: v0.24.3
+      - go.dependency:
+          name: sigs.k8s.io/structured-merge-diff/v4
+          lowerbound: v4.2.0
+          upperbound: v4.2.2
+- message: "dependency {{name}} with {{version}} is bad and you should feel bad for using it"
```
fabianvf, line 90: it doesn't look like this rule is actually triggering :\

fabianvf: /hold

engine/engine.go
fabianvf: good catch

rule-example.yaml
fabianvf: ah, the base image doesn't have maven installed so it returns 0 dependencies

provider/provider.go
fabianvf: we should probably change that because it's overloaded, and I think tracking where a dependency came from is probably something we'll want to do as well

fabianvf: https://github.com/konveyor/analyzer-lsp/issues/73

parser/rule_parser.go
```diff
@@ -378,18 +377,7 @@ func (r *RuleParser) getConditionForProvider(langProvider, capability string, va
...
-	}
-	found := false
-	for _, c := range providerCaps {
-		if c.Name == capability {
-			found = true
-			break
-		}
-	}
-	if !found {
+	if !client.HasCapability(capability) {
```
shawn-hurley, line 380: +1 on helper!

parser/rule_parser.go
```diff
@@ -400,7 +388,46 @@ func (r *RuleParser) getConditionForProvider(langProvider, capability string, va
 				ignorable = i
 			}
 		}
+	}
+
+	if capability == "dependency" {
```
shawn-hurley, line 393: Note to self's I think we may eventually want to refactor this method it has a ton of logic in it

shawn-hurley: Sorry not this one, but the one above, maybe I am just wrong too :)

provider/java/provider.go
```diff
@@ -85,13 +89,24 @@ func (p *javaProvider) Stop() {
...
-	return []lib.Capability{
+func (p *javaProvider) Capabilities() []lib.Capability {
+	caps := []lib.Capability{
 		{
 			Name:            "referenced",
 			TemplateContext: openapi3.SchemaRef{},
 		},
-	}, nil
+	}
+	if p.hasMaven {
```
shawn-hurley, line 99: +1

shawn-hurley (approved):
Looks good to me merge whenever!

PR #76, opened 2023-02-17
:sparkles: remove chain keyword order conditions

shawn-hurley:
First steps to making chaining of conditions more intuitive

1. Remove the chain keyword, Or and And do this now.
2. Ordering of from and as does not matter.

TODO: 
Adding a known struct for the output of a condition based on the incidents.
We will continue to have "extras" that will be unstructured.


fabianvf (approved):
Nice!

PR #80, opened 2023-02-28
:sparkles: add new category

pranavgaikwad:


shawn-hurley:
what is the difference between potential and optional?

pranavgaikwad:
@shawn-hurley so this is a category for violations that found a match for a certain condition but couldn't generate valid output (this is one example I found in windup report, there could be more) . For instance, an XML condition matched in an OR caused as a result of `builtin.xml` but the file was malformed and a valid violation couldn't be created for condition that followed `builtin.xml`

shawn-hurley:
I wonder if this is a second set of things that are errors? we already need to add a "ran but didn't find any issues" rules as well

pranavgaikwad:
@shawn-hurley makes sense to me. lets close this PR for now, I'll follow up with an issue instead

Issue #73, opened 2023-02-10
Track source as fileuri when reporting dependencies


fabianvf:
Currently we don't point to where the dependencies are coming from, which could be a useful piece of information. We should track and report that. We should also change the `Location` field in the `dependency.Dep` to match `Location` in the rest of the code base (ie it should point somewhere in the codebase) and find a different term to report what it currently gathers for java (or add it to an extras field for arbitrary metadata)

pranavgaikwad:
@fabianvf Did you just mean pointing to where the dependency is declared (pom.xml, go.mod) or you meant where it's actually used?

fabianvf:
Where it's declared

pranavgaikwad:
Location can be renamed to Type maybe?

PR #82, opened 2023-03-02
:sparkles: Adding SHA sums and renaming location to type

shawn-hurley:
fixes: #73


provider/java/dependency.go
```diff
@@ -37,8 +38,25 @@ func (p *javaProvider) GetDependencies() ([]dependency.Dep, error) {
...
+	}
+
+	// check errors
+	return string(outb.String())
+}
+
 func (p *javaProvider) GetDependenciesLinkedList() (map[dependency.Dep][]dependency.Dep, error) {
 
+	localRepoPath := p.getLocalRepoPath()
+	fmt.Printf("here: %v", localRepoPath)
```
pranavgaikwad, line 58: remove fmt.Printf?

pranavgaikwad (changes requested):

provider/java/dependency.go
shawn-hurley: good catch

provider/java/dependency.go
```diff
@@ -37,8 +38,24 @@ func (p *javaProvider) GetDependencies() ([]dependency.Dep, error) {
 	return deps, err
 }
 
+func (p *javaProvider) getLocalRepoPath() string {
+	cmd := exec.Command("mvn", "help:evaluate", "-Dexpression=settings.localRepository", "-q", "-DforceStdout")
+	var outb bytes.Buffer
+	cmd.Stdout = &outb
+	err := cmd.Run()
+	if err != nil {
+		fmt.Printf("here: %v", err)
```
pranavgaikwad, line 47: ~this guy too :)~

Actually I think we need this error reported right?

pranavgaikwad (approved):

provider/java/dependency.go
shawn-hurley: yeah, it probably should be a log but this is a good first issue we have these all over the place

PR #92, opened 2023-03-09
:book: Add Rules document

pranavgaikwad:


docs/rules.md
```diff
@@ -0,0 +1,265 @@
...
+```yaml
+when:
+  <conditions>
+```
+
+### When
+
+`When` consists of a condition.
+
+#### Provider Conditions
```
shawn-hurley, line 55: Do we want to add the provider settings docs here as well?

shawn-hurley (approved):

docs/rules.md
pranavgaikwad: @shawn I'll address that in a follow up

PR #99, opened 2023-03-16
:sparkles: Add jar files found to dependencies as well as handle default searches

shawn-hurley:
Must wait for https://github.com/konveyor/java-analyzer-bundle/pull/37 to merge.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -3,6 +3,7 @@ package provider
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/cbroglie/mustache"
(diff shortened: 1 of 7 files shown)


provider/provider.go
```diff
@@ -144,6 +145,7 @@ type DependencyCondition struct {
 	Upperbound string
 	Lowerbound string
 	Name       string
+	NameRegex  string
```
pranavgaikwad, line 148: Can we add an example rule for NameRegex?

pranavgaikwad (changes requested):
Changes look good to me, waiting for the bundles PR and also the test-output that's failing tests.

pranavgaikwad (approved):

PR #103, opened 2023-03-23
:sparkles: Add fallback dependency parser for Java

fabianvf:


Dockerfile
```diff
@@ -10,7 +10,7 @@ FROM quay.io/konveyor/jdtls-server-base
 WORKDIR /analyzer-lsp
 # TODO limit to prevent unnecessary rebuilds
 COPY  . /analyzer-lsp
-COPY --from=builder /analyzer-lsp/konveyor-analyzer /analyzer-lsp/konveyor-analyzer
+COPY --from=builder /analyzer-lsp/konveyor-analyzer /usr/bin/konveyor-analyzer
```
shawn-hurley, line 13: +1 this bit me in the demo :)

provider/provider.go
```diff
@@ -226,3 +223,23 @@ func (dc DependencyCondition) Evaluate(ctx context.Context, log logr.Logger, con
 
 	return resp, nil
 }
+
+func getVersion(depVersion string) (*version.Version, error) {
+	v, err := version.NewVersion(depVersion)
+	if err == nil {
+		return v, nil
+	}
+	// Parsing failed so we'll try to extract a version and parse that
```
shawn-hurley, line 232: for my edification, what are the cases this happens?

shawn-hurley (commented):
looks pretty good to me

provider/provider.go
jmle: Just in case you find it useful, this is what windup uses to compare versions: https://github.com/apache/maven/blob/master/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java#L538

used in here: https://github.com/windup/windup/blob/master/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/condition/Version.java#L56

shawn-hurley (approved):

provider/provider.go
fabianvf: yeah will do as a followup

PR #104, opened 2023-03-23
:sparkles:  Move tags output from violations to ruleset, take tags from a file

pranavgaikwad:
Adds ability to take tags from a file
Moves tag output to Ruleset level


pranavgaikwad:
@shawn-hurley not in the shim, we are requiring the tags file as an input to builtin provider which does the tag search.

shawn-hurley:
We search tags in the shim @fabianvf right? to make sure the tests are passing?

PR #115, opened 2023-04-11
:sparkles: Search for all files matching the pattern in the xml filepath

fabianvf:
This means if you look for `ejb-jar.xml` for example, it would also find `/config/ejb-jar.xml`, and allows you to put wildcards etc in the filepaths.

### DIFF
--- a/provider/builtin/provider.go
+++ b/provider/builtin/provider.go
@@ -185,6 +185,10 @@ func (p *builtinProvider) Evaluate(cap string, conditionInfo []byte) (lib.Provid
 			for _, pattern := range patterns {
 				files, err := findFilesMatchingPattern(p.config.Location, pattern)
 				if err != nil {
+					// Something went wrong dealing with the pattern, so we'll assume the user input
+					// is good and pass it on
+					// TODO(fabianvf): if we're ever hitting this for real we should investigate
+					fmt.Printf("Unable to resolve pattern '%s': %v", pattern, err)
 					xmlFiles = append(xmlFiles, pattern)
 				} else {
 					xmlFiles = append(xmlFiles, files...)


provider/builtin/provider.go
```diff
@@ -181,9 +181,24 @@ func (p *builtinProvider) Evaluate(cap string, conditionInfo []byte) (lib.Provid
 			xmlFiles = append(xmlFiles, xhtmlFiles...)
 		} else if len(cond.XML.Filepaths) == 1 {
 			// Currently, rendering will render a list as a space seperated paths as a single string.
-			xmlFiles = strings.Split(cond.XML.Filepaths[0], " ")
+			patterns := strings.Split(cond.XML.Filepaths[0], " ")
+			for _, pattern := range patterns {
+				files, err := findFilesMatchingPattern(p.config.Location, pattern)
+				if err != nil {
```
shawn-hurley, line 187: This bit seems odd, if there is an error then we add the pattern as a file? Can we add a comment on why?

shawn-hurley (approved):
I think I understand that he matching patterns will return an error if it is a file path, and I assume that is what we are doing with the if/else?

Overall I think I am fine adding pattern searching to file paths input.

provider/builtin/provider.go
fabianvf: my thinking was if something broke unintentionally it just goes back to the default behavior of trusting user input and running the rule

shawn-hurley: Yeah probably is the safest thing to do, I wonder can we add a debug log here I assume that we are going to be tracking down an obscure bug in the future :)

fabianvf: done

PR #118, opened 2023-04-18
:sparkles: add custom variables and code snippets 

### LINKED ISSUE #31: We need to figure out what File URI / URL is supposed to based on
We need to determine the root here and standardize the output

### LINKED ISSUE #117: Rule Authors Define Custom Extras
Right now, we have extras that can be used in the message for users, but a user may want to add their own.

this can be useful, in that a rule author now does not have to change the provider of the capability to eloquently write complex rules. We see examples of this use case in the eapX/websphere rules in windups rulesets.

shawn-hurley:
fixes #31 
fixes #117

### DIFF
--- a/demo-output.yaml
+++ b/demo-output.yaml
@@ -358,6 +358,7 @@
         message: found generic call
         codeSnip: element.get()
         variables:
+          VariableName: element
           file: file:///analyzer-lsp/examples/java/src/main/java/com/example/apps/App.java
           kind: Method
           lineNumber: 13
--- a/demo-output.yaml
+++ b/demo-output.yaml
@@ -7,121 +7,121 @@
     chain-pom-001:
       description: ""
       incidents:
-      - uri: file://examples/customers-tomcat-legacy/pom.xml
+      - uri: file:///analyzer-lsp/examples/customers-tomcat-legacy/pom.xml
         message: <groupId>com.fasterxml.jackson</groupId><artifactId>jackson-bom</artifactId><version>${jackson.version}</version><scope>import</scope><type>pom</type>
         extras:
           data: dependency
           innerText: "\n\t\t\t\tcom.fasterxml.jackson\n\t\t\t\tjackson-bom\n\t\t\t\t${jackson.version}\n\t\t\t\timport\n\t\t\t\tpom\n\t\t\t"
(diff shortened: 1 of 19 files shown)


hubapi/violations.go
```diff
@@ -60,9 +62,10 @@ type Violation struct {
...
 	// URI defines location in the codebase where violation is found
-	URI string `yaml:"uri"`
+	URI uri.URI `yaml:"uri"`
 	// Message text description about the incident
-	Message string `yaml:"message"`
+	Message  string `yaml:"message"`
+	CodeSnip string `yaml:"codeSnip,omitempty"`
 	// Extras reserved for additional data
 	//Extras json.RawMessage
 	Extras map[string]interface{}
```
shawn-hurley, line 71: Change this name to Variables too

shawn-hurley:
Still a Draft PR as I need to work on the custom variables, but the current implementation gives code snips.

engine/conditions.go
```diff
@@ -35,10 +37,36 @@ type ConditionEntry struct {
 }
 
 type IncidentContext struct {
-	FileURI string                 `yaml:"fileURI"`
-	Effort  *int                   `yaml:"effort"`
-	Extras  map[string]interface{} `yaml:"extras"`
-	Links   []hubapi.Link          `yaml:"externalLink"`
+	FileURI      uri.URI                `yaml:"fileURI"`
```
fabianvf, line 40: nice

fabianvf (approved):
looks good

demo-output.yaml
```diff
@@ -136,96 +136,66 @@
...
-      - uri: provider/lib/lib.go
+      - uri: file:///analyzer-lsp/provider/lib/lib.go
         message: all go files
-        extras: {}
-      - uri: provider/provider.go
+      - uri: file:///analyzer-lsp/provider/provider.go
         message: all go files
-        extras: {}
-      - uri: provider/provider_test.go
+      - uri: file:///analyzer-lsp/provider/provider_test.go
```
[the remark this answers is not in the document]
shawn-hurley: I have gone back and forth and I think we may need to different things for different reasons

I think provider to engine has to be abs path on filesystem. And then what we tell the hub can be relative to base of project. I wonder if we can just use the provider settings strip off the dir and say that becomes the base?

PR #119, opened 2023-04-21
:sparkles: Add support for xpath namespaces

fabianvf:


pranavgaikwad (changes requested):
LGTM. Would you add an example rule so we get output test coverage on new field?

shawn-hurley:
would the wind up rules test count?

fabianvf:
oh right I forgot about the test, my bad @pranavgaikwad I'll follow up with that

PR #134, opened 2023-05-02
:sparkles: when an invalid category, or no category is set, default it

### LINKED ISSUE #88: Consolidating Categories
## Current State
Today there are 4 Categories in the windup project.

1. Potential
2. Mandatory
3. Information
4. optional

The potential is supposed to mean that the Rule ran successfully, and there may or may not be an issue. We can't tell.

Mandatory is that the Rule ran correctly and the Rule states this must be fixed

Information is that the Rule ran correctly, and the Rule states that this is just for information, which could mean there is a problem in the future but not something you must fix.

Optional is that a rule ran unsuccessfully, and it might have a violation and may not.

## Proposed State

Consolidating these and making it easier for both end users to understand what the category is supposed to mean and what a rule writer should set is the goal of this issue. 

For the Rule's categories, I am proposing two mutually exclusive buckets. 

1. **Mandatory** will be when the Rule states that this must be fixed based on the ruleset. This will only be set for rules that have run successfully
2. **Potential** will be when the Rule states this is potentially an issue. This will only be set for rules that have run successfully

We need to create a new bucket for partial matches/errors. I would prefer to call this something like Errors, which will be different than the violations that use the categories above. In this world, we would have two new error categories:

1. Partial Match -> This Rule ran semi-successfully, and some conditions matched, but there were failures in others. This should include the error message and other stuff 
2. Failure -> this Rule did not run successfully in any way.

We also need a third bucket of rules for rules that ran successfully but did not have any violations. I believe this should just be something like PassedRules or something.

We should discuss the names here so that once we implement it, we can point back to this. 

Adding in folks that this will probably have an impact on:
@rromannissen @jortel @ibolton336 @pranavgaikwad @fabianvf

shawn-hurley:
Currently choosing to default to potential, as it has the least impact on the end user and captures the uncertainity of the default.

fixes #88

### DIFF
--- a/demo-output.yaml
+++ b/demo-output.yaml
@@ -6,6 +6,7 @@
   violations:
     chain-pom-001:
       description: ""
+      category: potential
       incidents:
       - uri: file:///analyzer-lsp/examples/customers-tomcat-legacy/pom.xml
         message: <groupId>com.fasterxml.jackson</groupId><artifactId>jackson-bom</artifactId><version>${jackson.version}</version><scope>import</scope><type>pom</type>
@@ -203,6 +204,7 @@
       effort: 3
     go-lang-ref-001:
       description: ""
+      category: potential
       incidents:
       - uri: file:///analyzer-lsp/examples/golang/main.go
         message: golang apiextensions/v1/customresourcedefinitions found file:///analyzer-lsp/examples/golang/main.go:9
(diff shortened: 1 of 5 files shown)


shawn-hurley:
@pranavgaikwad tagging rules should have no category thought IMO. Should we consider that we may need to handle anything w/ a message as having one of the three but only tagging rules must be nil? thoughts?

pranavgaikwad (approved):

djzager (approved):

PR #139, opened 2023-05-05
:sparkles: enable tag and message together

pranavgaikwad:


engine/engine.go
```diff
@@ -125,35 +125,48 @@ func (r *ruleEngine) createRuleSet(ruleSet RuleSet) *hubapi.RuleSet {
...
 					rule:        rule,
 					ruleSetName: ruleSet.Name,
 				})
 			} else {
-				metaRules = append(metaRules, ruleMessage{
+				taggingRules = append(taggingRules, ruleMessage{
 					rule:        rule,
 					ruleSetName: ruleSet.Name,
 				})
+				// if both message and tag are set
```
fabianvf, line 153: oh interesting, good idea, especially once we add a caching layer

fabianvf (commented):
Nice! I think there's a test we could leave in there but looks good to me

fabianvf (approved):
lgtm once conflict is fixed

PR #141, opened 2023-05-09
:sparkles: moving golang to external provider, hard coding values as first pass

### LINKED ISSUE #69: rework the all go files rule in demo testing
Everytime a go source file is added or removed it causes the demo test to fail. We should update that to be less specific so that it doesn't need to be constantly updated

shawn-hurley:
Adding GRPC proto, server/client libraries and writing wrappers around our interface's in golang to make converting easier. 

Fixes #69 
TODO:

1. Fix up provider_config to take the path to the correct server
2. Use "find_open_port"
3.


demo-output.yaml
```diff
@@ -212,29 +154,6 @@
           file: file:///analyzer-lsp/examples/golang/main.go
           lineNumber: 9
       extras: []
-    golang-gomod-dependencies:
```
shawn-hurley, line 215: Note that this is not there, because the dependency service is not added for external providers. 

I figure this could be a follow up

external-providers/golang-external-provider/go.mod
```diff
@@ -0,0 +1,44 @@
...
+	golang.org/x/net v0.8.0 // indirect
+	golang.org/x/sys v0.6.0 // indirect
+	golang.org/x/text v0.8.0 // indirect
+	google.golang.org/genproto v0.0.0-20230110181048-76db0878b65f // indirect
+	google.golang.org/grpc v1.54.0 // indirect
+	google.golang.org/protobuf v1.30.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace github.com/konveyor/analyzer-lsp => ../../
```
shawn-hurley, line 44: This will allow us to make changes quickly before we want to move this is actually out of tree and have to worry about releases and such

external-providers/golang-external-provider/main.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/bombsimon/logrusr/v3"
+	"github.com/konveyor/analyzer-lsp/provider"
+	"github.com/konveyor/analyzer-lsp/provider/lib"
+	"github.com/konveyor/golang-external-provider/pkg/golang"
+	"github.com/sirupsen/logrus"
+)
+
+func main() {
+	configs, err := lib.GetConfig("/analyzer-lsp/provider_settings.json")
```
shawn-hurley, line 15: This is hardcoded right now, to work in the container images. We need to re-do providers settings into two things:

1. Settings that the provider has that are hard coded
2. Settings to the analyzer binary that tells it where to look we have to figure out if we need some sort of expansion to the protocol to deal with a "session"

testing.Dockerfile
```diff
@@ -1,12 +1,21 @@
+FROM golang:1.18 as builder
```
shawn-hurley, line 1: These changes allows us to fix the golang tests that we have had on our backlog for a while

shawn-hurley (commented):
Adding comments for places where I know I cut corners

Dockerfile
```diff
@@ -11,7 +11,7 @@ WORKDIR /analyzer-lsp
 # TODO limit to prevent unnecessary rebuilds
 COPY  . /analyzer-lsp
 COPY --from=builder /analyzer-lsp/konveyor-analyzer /usr/bin/konveyor-analyzer
+COPY --from=builder /analyzer-lsp/external-providers/golang-external-provider/golang-external-provider /usr/bin/golang-external-provider
 COPY provider_container_settings.json /analyzer-lsp/provider_settings.json
 
-
-CMD ["/bin/bash", "-c", "go run main.go;cat output.yaml"]
+CMD ["/bin/bash", "-c", "golang-external-provider & disown -r; go run main.go; cat output.yaml"]
```
fabianvf, line 17: interesting, I assumed the plugins would be given to the engine and it would handle starting them up as part of the init process. I think this is fine but maybe a little weird from a UX perspective? I don't love requiring the users to start up all the providers by hand

demo-output.yaml
fabianvf: was there additional difficulty moving that interface over, or did you just want to keep the PR smaller?

engine/engine.go
```diff
@@ -168,6 +168,7 @@ func (r *ruleEngine) RunRules(ctx context.Context, ruleSets []RuleSet) []hubapi.
 					r.logger.Info("rule returned", "rule", response)
 					defer wg.Done()
 					if response.Err != nil {
+						fmt.Printf("\n%v\n", response.Err)
```
fabianvf, line 171: is this on purpose?

external-providers/golang-external-provider/main.go
fabianvf: I think users should only need to provide one config file to the engine, it should handle forwarding settings to the providers imo

external-providers/golang-external-provider/pkg/golang/provider.go
```diff
@@ -17,6 +17,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+// TODO(shawn-hurley): Pipe the logger through
+// Determine how and where external providers will
```
fabianvf, line 21: will what?

external-providers/golang-external-provider/provider-settings.yaml
```diff
@@ -0,0 +1,7 @@
+[
```
fabianvf, line 1: why would it have a separate config from the engine?

main.go
```diff
@@ -85,6 +85,11 @@ func main() {
 			os.Exit(1)
 		}
 		providers[config.Name] = provider
+		err = provider.Init(ctx, log)
+		if err != nil {
+			log.Error(err, "unable to create provider client")
```
fabianvf, line 90: we should mention which provider so the user can fix it

provider/grpc/provider.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	return lib.ProviderEvaluateResponse{
+		Matched:         true,
+		Incidents:       incs,
+		TemplateContext: r.Response.TemplateContext.AsMap(),
+	}, nil
+}
+
+// We don't have dependencies
```
fabianvf, line 150: These are TODOs though, right?

provider/grpc/provider.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+type grpcProvider struct {
+	Client pb.ProviderServiceClient
+	log    logr.Logger
+	ctx    context.Context
+	conn   *grpc.ClientConn
+}
+
+func NewGRPCClient() *grpcProvider {
```
fabianvf, line 25: Should the connection string not be an input to this?

provider/provider.go
```diff
@@ -143,12 +142,12 @@ func templateCondition(condition []byte, ctx map[string]lib.ChainTemplate) ([]by
...
 func GetProviderClient(config lib.Config) (Client, error) {
 	switch config.Name {
-	case "go":
-		return golang.NewGolangProvider(config), nil
 	case "java":
 		return java.NewJavaProvider(config), nil
 	case "builtin":
 		return builtin.NewBuiltinProvider(config), nil
+	case "go":
+		return grpc.NewGRPCClient(), nil
```
fabianvf, line 150: I don't understand how we would break out e.g. the Java provider without this being parameterized in some way

fabianvf (commented):
The structure of running the provider separately from the main process is strange to me, my assumption was the main process would take in all the configs, spin up the required providers and use grpc to communicate with them. I think from the users perspective the experience should feel like a single binary, even if there are multiple binaries on the system being executed to achieve the analysis

provider/grpc/provider.go
shawn-hurley: It will, but for first pass, didn't handle breaking up the config

provider/grpc/provider.go
shawn-hurley: Yes 100%

engine/engine.go
shawn-hurley: Nope cleanup needed +1

external-providers/golang-external-provider/pkg/golang/provider.go
shawn-hurley: Honestly don't remember  👍🏻 

Will try and update on next pass.

main.go
shawn-hurley: +1

provider/grpc/provider.go
shawn-hurley: captured in #142 Resolving

provider/provider.go
shawn-hurley: Captured in #142 resolving this for now

Dockerfile
shawn-hurley: Discussion and update incoming captured in #142

external-providers/golang-external-provider/main.go
shawn-hurley: Captured in #142 I agree

demo-output.yaml
shawn-hurley: I was trying to keep it small but might as well do this as well.

pranavgaikwad (changes requested):
seeking answers to two questions...

pranavgaikwad (approved):
I created https://github.com/konveyor/analyzer-lsp/issues/158 to follow up on one question I had. but looks good otherwise.

provider/provider.go
```diff
@@ -2,38 +2,181 @@ package provider
...
+type Capability struct {
+	Name            string
+	TemplateContext openapi3.SchemaRef
+}
+
+type Config struct {
+	Name       string     `yaml:"name,omitempty"`
+	BinaryPath string     `yaml:"binaryPath,omitempty"`
+	Address    string     `yaml:"address,omitempty"`
+	InitConfig InitConfig `yaml:"initConfig,omitempty"`
```
shawn-hurley, line 45: TODO: Add Issue for multiple Init Configs

provider/provider.go
```diff
@@ -2,38 +2,181 @@ package provider
...
+	Location string `yaml:"location,omitempty"`
+
+	// This is the path to look for the dependencies for the project.
+	// It is relative to the Location
+	DependencyPath string `yaml:"dependencyPath,omitempty"`
+
+	LSPServerPath string `yaml:"lspServerPath,omitempty"`
+
+	// This will have to be defined for each provider
+	ProviderSpecificConfig map[string]string `yaml:"providerSpecificConfig,omitempty"`
```
shawn-hurley, line 60: TODO: Move to map[string]interfaace{}

provider/provider.go
shawn-hurley: created #158  

I think this is going to cause use some more changes, can we do this as a follow-up?

@pranavgaikwad @fabianvf

provider/provider.go
shawn-hurley: updated.

Dockerfile
```diff
@@ -11,7 +11,7 @@ WORKDIR /analyzer-lsp
 # TODO limit to prevent unnecessary rebuilds
 COPY  . /analyzer-lsp
 COPY --from=builder /analyzer-lsp/konveyor-analyzer /usr/bin/konveyor-analyzer
+COPY --from=builder /analyzer-lsp/external-providers/golang-external-provider/golang-external-provider /usr/bin/golang-external-provider
 COPY provider_container_settings.json /analyzer-lsp/provider_settings.json
 
-
-CMD ["/bin/bash", "-c", "go run main.go;cat output.yaml"]
+CMD ["/bin/bash", "-c", "golang-external-provider & disown && sleep 5; konveyor-analyzer --output-file output.yaml; cat output.yaml"]
```
fabianvf, line 17: Do we still need to do this?

shawn-hurley: I think I have the config set to address because we can not manage the ports yet probably need to add a follow-up?

shawn-hurley: https://github.com/konveyor/analyzer-lsp/issues/166

fabianvf: sounds good to me

fabianvf (approved):

PR #146, opened 2023-05-11
:sparkles: implement label selector for rules

### LINKED ISSUE #130: Support a general purpose selector for rule labels
We need the ability to include / exclude rules based on labels. We think it'd be beneficial to have a general purpose label selector that supports AND, OR and NOT operation on labels so that we can support complex filtering for potential future use cases, while also satisfying needs for ongoing windup integration.

pranavgaikwad:
Once this is merged, we should follow up with shim PR to add labels on rules.

fixes #130

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -141,6 +141,7 @@ func (r *ruleEngine) RunRules(ctx context.Context, ruleSets []RuleSet, selectors
 		for _, rule := range ruleSet.Rules {
 			// skip rule when doesn't match any selector
 			if !matchesAllSelectors(rule.RuleMeta, selectors...) {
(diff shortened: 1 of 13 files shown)


engine/engine.go
```diff
@@ -422,3 +429,12 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 func (r *ruleEngine) createPerformString(messageTemplate string, ctx map[string]interface{}) (string, error) {
 	return mustache.Render(messageTemplate, ctx)
 }
+
+func matchesAllSelectors(m RuleMeta, selectors ...RuleSelector) bool {
```
shawn-hurley, line 433: Can we comment this with why it must match all selectors

shawn-hurley (commented):
LGTM besides the comment but makes a ton of sense

shawn-hurley (approved):

Issue #144, opened 2023-05-10
Deduplication of Incidents Per rule


shawn-hurley:
If a rule has an incident that has the exact same file/line number, we should only keep one of these, currently, we keep all of them.

shawn-hurley:
To implement this, when creating violations, we should have a map of URI-linenumber, that we can say to ignore and skip creating violations for things that have already been found.

Chanakya-TS:
incidents can have the same uri but different message, line number etc. cannot have a map of uri, need to be able to compare incident structs to stop duplicates

shawn-hurley:
The serialization cost is probably expensive I wonder if there is a way to compare just URI and resolved message

pranavgaikwad:
`fmt.Sprintf("%s-%s-%s", fileURI, messageHash, lineNumber)` isn't too bad for the map key

Chanakya-TS:
there are incidents with minute differences such as the following one, so just fileURI, messageHash and lineNumber wont be sufficient.

![Image](https://github.com/konveyor/analyzer-lsp/assets/66557279/73dd6445-a0ae-4022-b5e1-c209fe93e91c)

shawn-hurley:
I might have a different opinion than others, but if that is found by the same rule, I think only one of them being a violation is fine IMOOn May 17, 2023, at 5:22 PM, Chanakya-TS ***@***.***> wrote:﻿
there are incidents with minute differences such as the following one, so just fileURI, messageHash and lineNumber wont be sufficient.

PR #161, opened 2023-05-17
:bug: stopped duplication of incidents while creating violations

Chanakya-TS:
fixes issue #144


engine/engine.go
```diff
@@ -416,7 +418,17 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 			incident.Message = templateString
 		}
 
-		incidents = append(incidents, incident)
+		// Checking duplication of incident
+
+		jsonIncident, _ := json.Marshal(incident)  // Converting to JSON to be used as keys of a map
```
shawn-hurley, line 423: I am worried about the cost of this. Can we make a key function that only takes specific values to compare? 

Think like file URI, Message, and Linenumber if it exists in extras. I think that the cost of marshal could become too great, and the things that this is capturing as different issues, I think is misleading. 

For instance, if the type is a method or a local variable, as long as we are pointing to the same location, it is finding the same violation, just in a different search mechanism, and I think the user can be spared this detail IMO.

engine/engine.go
```diff
@@ -416,7 +418,17 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
...
+		stringJsonIncident := string(jsonIncident) // Converting JSON to string
+
+		_, isDuplicate := incidentsSet[stringJsonIncident] // Checking if incident already exists in set
```
shawn-hurley, line 426: I usually perfer to make this a one line, but this is just a nit.

```
if _, ok := ...... ; !ok {
  do x
  }
  ```

Comment (Chanakya-TS) [reply engine/engine.go:426]: > I usually perfer to make this a one line, but this is just a nit.
> 
> ```
> if _, ok := ...... ; !ok {
>   do x
>   }
> ```

ah okay, will do that from next time. thank you!

**Review by shawn-hurley — COMMENTED** [review]
I think this is a great PR, I do think that we may need to update the rule-example.

I believe that this is filtering out (now correctly 😄 ) the XML XPath on the pom file. 

One way to fix this is to update the `rule-example.yaml` to include a variable in the message.

**Review by pranavgaikwad — CHANGES_REQUESTED** [review]
Our CI expects content of the `container-output.yaml` to match the output after running analyzer with your changes. It looks like the change you made deduplicates a lot of the incidents our CI is expecting. Can you please  update the `container-output.yaml` to match the new expected output?

File: demo-output.yaml [hunk 3]
```diff
@@ -138,8 +138,76 @@
       - testing
       - test
       incidents:
+      - uri: file:///analyzer-lsp/cmd/analyzer/main.go
```
shawn-hurley, line 141: These should not be added back, we got rid of these

Chanakya-TS: thats weird, gimme a sec

Chanakya-TS: i think something happened while rebasing

Chanakya-TS: wait no, that is the output that i am getting when i run the build. im not sure what the problem is

shawn-hurley (approved):
I think this looks good I wonder if it will kill our integration numbers

pranavgaikwad (approved):

pranavgaikwad:
I was just making sure we don't break integration test with this, and I found https://github.com/konveyor/analyzer-lsp/blob/2f7088e1a51b8ea40b7343fff679d43a972b9b44/provider/internal/builtin/provider.go#L187, I think we need safeguard for a non integer type line number

PR #192, opened 2023-06-01
Allow parentheses in tags

pranavgaikwad:
Noticed some rules failing because of this in shim


shawn-hurley (commented):
LGTM thanks for the test!

PR #195, opened 2023-06-05
:bug: pulling line number out of variables in incident struct

### LINKED ISSUE #186: Pull up `lineNumber` from `Extras` to `Incident`

Chanakya-TS:
fixes #186


Chanakya-TS:
currently, i am pulling the line number out of variables and putting it separately in the incident. im not sure if this is what is required. i may have to query the provider the line number and pipe it through for every incident but i am not sure how to do that. 

i also need to figure out how to make sure that the {{lineNumber}} in rules message takes the line number which is now outside the variables.

engine/engine.go
```diff
@@ -357,9 +357,16 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 		if r.incidentLimit != 0 && len(incidents) == r.incidentLimit {
 			break
 		}
+		lineNumber, hasLineNumber := m.Variables["lineNumber"]
```
shawn-hurley, line 360: I think that is reasonable to update the contract of the providers to include this as a top level. I would look 
https://github.com/konveyor/analyzer-lsp/blob/main/provider/provider.go#L117 to start.

shawn-hurley (changes requested):
Good start, but I think that you can go even further down the stack and force the providers to choose. 

Also a minor comment on the API. 

This is a really good start though!

shawn-hurley:
I would mark this reviewable and i would update the demo-output but otherwise this looks good. Sonds like the pointer will be a followup. I am ok with it as long as it is a quick follow on. @Chanakya-TS can you create an issue and align it to Demo E?

shawn-hurley (approved):
This looks great! move it out of draft so we can merge!

PR #196, opened 2023-06-05
:bug: changes the output format for dependencies 

### LINKED ISSUE #140: Change dependency output into a list by provider
Right now, the dependencies from all providers are output in the same list or tree. 

For instance, if `java` provider provides `dep2` and `dep3`, while the `go` provider gives `dep1`. The current output looks like:

```yaml
- dep1
- dep2
- dep3
```  

We should separate them by providers like so: 

```yaml
- provider: java
  dependencies:
  - dep2
  - dep3
- provider: golang
  dependencies:
  - dep1
```

This will also help populate the language field in the UI

Chanakya-TS:
fixes #140

### DIFF
--- a/cmd/dep/main.go
+++ b/cmd/dep/main.go
@@ -20,13 +20,13 @@ var (
 )
 
 type DepsFlatItem struct {
-	Provider     string         `yaml:"Provider"`
-	Dependencies []provider.Dep `yaml:"Dependencies"`
+	Provider     string         `yaml:"provider"`
+	Dependencies []provider.Dep `yaml:"dependencies"`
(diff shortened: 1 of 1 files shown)


cmd/dep/main.go
```diff
@@ -47,34 +58,46 @@ func main() {
 			log.Error(err, "unable to create provider client")
 			os.Exit(1)
 		}
+		err = provider.ProviderInit(context.TODO())
```
pranavgaikwad, line 61: good catch

pranavgaikwad (approved):

PR #204, opened 2023-06-12
:bug: update demo rule

### LINKED ISSUE #110: Mistake in rule-example?:  apiextensions/v1beta1/customresourcedefinitions is deprecated, apiextensions/v1/customresourcedefinitions should be used instead
Think we might have a mistake with the below example on line 35

https://github.com/konveyor/analyzer-lsp/blob/main/rule-example.yaml#L35

```
- message: apiextensions/v1beta1/customresourcedefinitions is deprecated, apiextensions/v1/customresourcedefinitions should be used instead
  ruleID: lang-ref-001
  when:
    or:
    - java.referenced:
        pattern: "*CustomResourceDefinition*"
        location: TYPE
    - go.referenced: "*CustomResourceDefinition*"
```

Assume this will match for all uses of CustomResourceDefinition, not just the apiextensions/v1beta1 the message states?

pranavgaikwad:
Fixes #110

### DIFF
--- a/demo-output.yaml
+++ b/demo-output.yaml
@@ -210,8 +210,8 @@
         lineNumber: 2
         variables:
           file: file:///analyzer-lsp/examples/java/src/main/java/com/example/apps/App.java
-          kind: Module
-          name: io.fabric8.kubernetes.api.model.apiextensions.v1beta1.CustomResourceDefinition
+          kind: Method
+          name: main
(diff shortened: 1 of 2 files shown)


pranavgaikwad:
I just discovered that the lineNumber is not working as expected for the go provider, creating an issue.

shawn-hurley (approved):
Looks good to me; does need a rebase though

PR #210, opened 2023-06-14
:bug: Adding the deps to the external package for the apis

shawn-hurley:
I was moving things around based on a convo with @jortel so that only one package has to be used by external folks.

### DIFF
--- a/cmd/dep/main.go
+++ b/cmd/dep/main.go
@@ -7,6 +7,7 @@ import (
 	"os"
 
 	"github.com/bombsimon/logrusr/v3"
+	"github.com/konveyor/analyzer-lsp/hubapi"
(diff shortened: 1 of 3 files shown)


pranavgaikwad (changes requested):
@shawn-hurley I think we need these structs in there too https://github.com/konveyor/analyzer-lsp/blob/756cd8e39cf8068896533e86c8d9e00daa56d1fc/cmd/dep/main.go#L22-L30

I'd actually put them both under one type, and only set depTree or depPath, leaving other omitempty

shawn-hurley:
@pranavgaikwad updated, what do you think of adding the URI that we need to differentiate between init configs here?

pranavgaikwad (approved):
Thank you for working on this Shawn

PR #211, opened 2023-06-15
:bug: Update filter code to use konveyor-jdt

pranavgaikwad:
Making it a constant so that we can use that in the bundle if we wanted to


shawn-hurley (approved):
Merge this asap, sorry about missing this; thanks for digging in and fixing @pranavgaikwad

Issue #212, opened 2023-06-15
Add dependency binary to test so we don't break the basic dep flow


pranavgaikwad:
We need a basic test for dependency CLI as well

PR #213, opened 2023-06-15
:bug: Fix dependency CLI and add it to Dockerfile

pranavgaikwad:
Follow  up with https://github.com/konveyor/analyzer-lsp/issues/212


shawn-hurley (approved):
+1 on this sorry about missing this. 

I think that we should add to testing as well, which I think you are working on too. once that is is I am plus one to merge

PR #215, opened 2023-06-16
:bug: add dependency to test, use multiple init configs in demo

### LINKED ISSUE #206: Use multiple init config in building provider in the testing Dockerfile
We need to make sure that the multiple init thing is tested.

### LINKED ISSUE #212: Add dependency binary to test so we don't break the basic dep flow

pranavgaikwad:
Fixes #212 
Fixes #206

### DIFF
--- a/demo-output.yaml
+++ b/demo-output.yaml
@@ -10,6 +10,25 @@
       description: ""
       category: potential
       incidents:
+      - uri: file:///analyzer-lsp/examples/java/pom.xml
+        message: <groupId>junit</groupId><artifactId>junit</artifactId><version>4.11</version><scope>test</scope>
+        variables:
+          data: dependency
+          innerText: "\n      junit\n      junit\n      4.11\n      test\n    "
+          matchingXML: <groupId>junit</groupId><artifactId>junit</artifactId><version>4.11</version><scope>test</scope>
+      - uri: file:///analyzer-lsp/examples/java/pom.xml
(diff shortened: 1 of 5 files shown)


pranavgaikwad:
this change should be failing the test, investigating why its not

demo-output.yaml
```diff
@@ -10,6 +10,25 @@
       description: ""
       category: potential
       incidents:
+      - uri: file:///analyzer-lsp/examples/java/pom.xml
```
pranavgaikwad, line 13: the order of this changed after I updated the init config

provider/internal/builtin/service_client.go
```diff
@@ -61,7 +61,7 @@ func (p *builtintServiceClient) Evaluate(cap string, conditionInfo []byte) (prov
 				continue
 
 			}
-			ab, err := filepath.Abs(filepath.Join(p.config.Location, match))
+			ab, err := filepath.Abs(match)
```
pranavgaikwad, line 64: `walkDir` is already appending the base location to the matching filenames because we are passing it as root path. We never found this because location for builtin was empty. I still need to test this with shim once just to confirm it doesn't break anything else.

pranavgaikwad: looking good with the shim tests

```
Overall success rate: 80.75% (1909/2364)
```

shawn-hurley:
Big +1 and lgtm

shawn-hurley (approved):

PR #218, opened 2023-06-16
:sparkles: Renaming output package and updating for file per list of deps

### LINKED ISSUE #225: Handle file URIs for dependency identified from a JAR file better
Right now, JAR files are simply associated with the POM file of the project. This is problematic in case of archives where we don't necessarily find the right pom file at project root. OR in general, it's better to point to the JAR path itself.

shawn-hurley:
Fixes #225


provider/internal/java/dependency.go
```diff
@@ -152,7 +158,10 @@ func (p *javaServiceClient) GetDependenciesDAG() ([]provider.DepDAGItem, uri.URI
 	}
 	filepath.WalkDir(moddir, w.walkDirForJar)
```
pranavgaikwad, line 159: I think all .jar files here will be incorrectly associated with the pom file `file`. This was always a problem I think. I am fine with addressing this later though. We need two fixes here: 1. Change file URI for JARs 2. Look for JARs in fallback parser too

pranavgaikwad (approved):
LGTM, thanks for working on this! I have one question related to JARs

provider/internal/java/dependency.go
shawn-hurley: Yeah I agree I can make this change fairly quickly I think so I think we should just do it on this PR

pranavgaikwad: @shawn-hurley thanks, I had already created an issue to capture that work, I will link it with the PR

shawn-hurley: Perfect pretty sure the change should be there now

PR #228, opened 2023-06-21
:sparkles: added helper fn to filter on filepath and regex 

krishnaduttPanchagnula:


shawn-hurley:
Adding some extra clarity for reviewers. We were supposed to work #228 but I screwed up and instead wrote down instructions for a differnt helper function for providers, the ability to filter out results based on a FilePattern. 

Adding @pranavgaikwad and @fabianvf for reviews 

@krishnaduttPanchagnula Thank you for the patience and I am hoping that we can continue working together it was nice meeting!

fabianvf (approved):
A test would be nice, not a blocker though

pranavgaikwad:
@krishnaduttPanchagnula Thank you for submitting a PR!

PR #230, opened 2023-06-23
:sparkles: enable taking http proxy settings

### LINKED ISSUE #101: Analyzer CLI should be able to take HTTP / HTTPS proxies for maven

pranavgaikwad:
Fixes #101 

1. Adds a new `proxyConfig` in provider settings. This is so that providers know whether a proxy should be used. This is technically only useful  for  external providers. Right now, we only have go provider which is basically taking the provider settings and setting proxy config as environment variables because `go` respects standard system settings. Other providers, however, may use these settings differently. 
2. When a specific config is not specified, system-wide proxy settings set via env vars are respected. Everything running in the main process respects system-wide settings, thanks to golang.
3. Maven running within JDTLS, however, does not respect system settings. This should be a bug IMO because `java.net.useSystemProxy` is a env var that most Java programs respect, which m2e core doesn't seem to respect. For this case, we need to specify settings via `settings.xml` file. This file is a common practice within users to set proxy settings for maven, we also have support for taking a settings file via Hub. As a result, we are adding a `providerSpecificConfig` field for Java that allows taking this file in addition to the normal proxy settings.

*NOTE*: There is one case which I haven't tested - when an external provider runs outside the network of the main process, and there is a proxy to establish connection between the  provider and the rule engine itself. I believe that gRPC client should respect system proxies in that case too and we don't need to do anything special there. I am not sure right now. I am creating an issue just to make sure we don't lose track  of that case.


shawn-hurley (approved):
Great PR, this makes a ton of sense to me.

I do wondering if we can capture as a follow-up to have some CI that is able to test this code path. I think it is potentially really easy to break this and we would never know.

pranavgaikwad:
needs rebase

PR #233, opened 2023-06-27
moved xml logic to lib.go

### LINKED ISSUE #187: Allow all builtin types to handle file paths from files
Currently, only XML can handle file paths from the `files` condition.

We should make this a library function such that other providers can also determine if the field is in the template, and if so, automatically pull out the file paths and give them an `[]string` 

You can find the file condition code here: https://github.com/konveyor/analyzer-lsp/blob/main/provider/internal/builtin/service_client.go#L75-L127

you can see the example of XML: https://github.com/konveyor/analyzer-lsp/blob/main/provider/internal/builtin/service_client.go#L135-L145

I believe that we should consider the library function living in provider.go here:
https://github.com/konveyor/analyzer-lsp/blob/main/provider/provider.go

krishnaduttPanchagnula:
Closes #187 

Changes:
 - Moved the FindFilesMatchingPattern from service_client.go to lib.go ( in provider package) and made the function public
 - Converted the xml case condition as a helper function in lib.go


krishnaduttPanchagnula:
Hi @shawn-hurley i have raised the PR based on our discussion. Please review it and suggest which icon should i add to this PR.

pranavgaikwad:
@krishnaduttPanchagnula Thank you for submitting a PR! Please look at the failing test in "Demo Testing" CI test. The  `rule-example.yaml` file at repo root is what is used as input rules when running the CI and expected output must be `demo-output.yaml`. It looks like this PR is causing a change in that output, please take a look at what's changed and see if its expected or not. Thank you again!

provider/internal/builtin/service_client.go
```diff
@@ -132,52 +130,14 @@ func (p *builtintServiceClient) Evaluate(cap string, conditionInfo []byte) (prov
 			return response, fmt.Errorf("Could not parse provided xpath query '%s': %v", cond.XML.XPath, err)
 		}
 		//TODO(fabianvf): how should we scope the files searched here?
-		var xmlFiles []string
-		if len(cond.XML.Filepaths) == 0 {
```
fabianvf, line 136: I think we want this branch to remain in the provider, since it's pretty specific to what the XML provider needs and isn't generally useful. The `else if` and `else` branches make sense to move out though

provider/lib.go
```diff
@@ -17,3 +21,74 @@ func FilterFilePattern(regex string, filepath string) (bool, error) {
...
+		if matched {
+			matches = append(matches, path)
+		}
+		return nil
+	})
+	return matches, err
+}
+
+func GetFiles(filePaths []string, rootpath string) ([]string, error) {
+	var xmlFiles []string
```
fabianvf, line 44: any references to XML files should be removed/renamed from this function, since this is a generic function for getting files of any kind.

fabianvf (changes requested):

provider/internal/builtin/service_client.go
```diff
@@ -280,30 +240,3 @@ func (p *builtintServiceClient) Evaluate(cap string, conditionInfo []byte) (prov
...
-	err := filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
-		if err != nil {
-			return err
-		}
-		var matched bool
-		if regex != nil {
-			matched = regex.MatchString(d.Name())
-		} else {
-			// TODO(fabianvf): is a fileglob style pattern sufficient or do we need regexes?
-			matched, err = filepath.Match(pattern, d.Name())
```
fabianvf, line 298: I think the error in the demo output is coming from the removal of this branch, `*.go` is a fileglob style pattern instead of regex and so will not compile.

krishnaduttPanchagnula: implemented

krishnaduttPanchagnula:
Hi @fabianvf  i have implemented the changes that you asked. Can you please review

pranavgaikwad:
Hey all, please forgive my ignorance here...I don't see how this solves #187 ? Or am I missing something?

provider/internal/builtin/service_client.go
```diff
@@ -135,89 +135,67 @@ func (p *builtintServiceClient) Evaluate(cap string, conditionInfo []byte) (prov
 		var xmlFiles []string
 		if len(cond.XML.Filepaths) == 0 {
```
shawn-hurley, line 136: Where and how, do we handle when there are filepaths that are passed in?

provider/lib.go
```diff
@@ -17,3 +21,62 @@ func FilterFilePattern(regex string, filepath string) (bool, error) {
...
+		}
+		return nil
+	})
+	return matches, err
+}
+
+func Getfiles(configLocation string, filepaths []string) ([]string, error) {
+	var xmlFiles []string
+
+	if len(filepaths) == 1 {
```
shawn-hurley, line 55: I am having a tough time seeing why a single filepath would be different here.

shawn-hurley (changes requested):
I think that GetFiles has to handle all the logic of dealing with filepaths and defaulting by matching patterns.

to do this, make GetFiles take the filepaths, and a varidic of patterns. if filepaths, then find all the absolute filepaths for that. if there are no patterns,then return empty list of files. if no filepaths and patterns, then defautl the list to the combonation of all the files found from the patterns 

once you do this, add filepaths to JSON, and add a test that proves that you can use filepaths and it still defautls to looking for all the json files.

provider/lib.go
fabianvf: what do you mean different?

shawn-hurley: I forgot about the edge case that we talked about the other day, where when there is a single file path, it could be that it is the rendered output from the template, and it could be the case that it is a space-separated list.

I think that we can resolve this one, my bad folks

provider/internal/builtin/service_client.go
fabianvf: yeah, looks like we're missing that branch. GetFiles (https://github.com/konveyor/analyzer-lsp/pull/233/files/f958037f3de788a7c1a90abe15cfed88b9558c05#diff-6335e3249b8d9e8dfe8312aacc749597539e4c377dc0fc779ab68c224fa25f9cR147) is being called with a cond.XML.FilePaths that we already know is empty, we need to move that to an `else` branch for populating `xmlFiles`

shawn-hurley (approved):
I think this looks good to me!

pranavgaikwad:
Testing with shim now

pranavgaikwad:
OK, I tested with shim, and looking good to merge. Thank you!

shawn-hurley:
Thank you for the Contributions!

PR #240, opened 2023-06-30
:sparkles: Include line number in snip

### LINKED ISSUE #200: Add line numbers to code snip
Refer to https://github.com/konveyor/enhancements/issues/124

fabianvf:
Also disable automatic 80 character line splits from the yaml library.

Closes #200

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


fabianvf:
I'm pretty sure this is the right format for the UI but could use a doublecheck

engine/engine.go
```diff
@@ -499,11 +499,11 @@ func (r *ruleEngine) getCodeLocation(m IncidentContext, rule Rule) (codeSnip str
 		codeSnip := ""
 		for scanner.Scan() {
 			if (lineNumber - CONTEXT_LINES) == m.CodeLocation.EndPosition.Line {
-				codeSnip = codeSnip + fmt.Sprintf("%v", scanner.Text())
+				codeSnip = codeSnip + fmt.Sprintf("%d  %v", lineNumber+1, scanner.Text())
 				break
 			}
 			if (lineNumber + CONTEXT_LINES) >= m.CodeLocation.StartPosition.Line {
-				codeSnip = codeSnip + fmt.Sprintf("%v", scanner.Text())
+				codeSnip = codeSnip + fmt.Sprintf("%d  %v\n", lineNumber+1, scanner.Text())
```
pranavgaikwad, line 506: @fabianvf UI is expecting line numbers to be right-aligned. Can we use `%*d  %v\n` and pass the width as a parameter to Sprintf() based on the maximum line number we see. That will format the numbers like below:
```bash
 8
 9
10
```

pranavgaikwad (changes requested):

engine/engine.go
fabianvf: @pranavgaikwad I can do that, though it seems strange to put UI formatting logic into the backend like this IMO

pranavgaikwad: @fabianvf yup that's true, but I referred to a discussion on this topic here https://github.com/konveyor/enhancements/issues/124#issuecomment-1599310182 and it seems UI is expecting that right now. I am fine with merging this as-is and addressing that in the UI

pranavgaikwad (approved):

PR #243, opened 2023-07-03
:bug: allow dots in label values

### LINKED ISSUE #242: Rule label selector should not fail on dots in label values
This is an issue right now. Our current language uses `.` operator as a token separator, and we will need to override that if we were to allow dots in the values. A simple solution here would be to get rid of dots from labels altogether.

pranavgaikwad:
Fixes #242

### DIFF
--- a/engine/labels/labels.go
+++ b/engine/labels/labels.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


fabianvf (approved):
Nice! Glad it wasn't the language parser

PR #252, opened 2023-07-11
:bug: Strip line numbers from code snips before applying substring matches :bell: 

fabianvf:
This allows the substrings to successfully match instead of breaking them completely. Janky solution though, may be better to expand the CodeSnip interface to allow this more easily?


pranavgaikwad (approved):

pranavgaikwad:
I am good with this change

PR #253, opened 2023-07-12
:sparkles: add code snip to filecontent

pranavgaikwad:


pranavgaikwad:
We eventually need to do this for XML and JSON too #254

Issue #262, opened 2023-07-17
Add helper functions to manipulate labels


pranavgaikwad:
There are several places in the source code where we repeatedly manipulate Label strings. The labels have a well known format, we should refactor and introduce some helper functions to do label manipulation. 

An example: 

https://github.com/konveyor/analyzer-lsp/blob/357b17b9472fe365bebafeb70691d8bf71220e3f/engine/labels/labels.go#L195

If you search for `"%s=` in the source code you'll find more occurrences.

PR #265, opened 2023-07-19
:sparkles: Refactor Label Manipulation

RipulHandoo:
Pull Request: Refactor Label Manipulation

Description:
Introducing the 'formatLabel' helper function to refactor and centralize key-value label manipulation throughout the codebase. This change addresses the issue of repeated label string manipulation, enhancing code organization and maintainability.

Changes Made:
- Added 'formatLabel' function to handle label key-value formatting.
- Replaced instances of label construction with calls to 'formatLabel'.
- Ensured consistent usage of 'formatLabel' for label manipulation.

This pull request resolves issue #262

### DIFF
--- a/engine/labels/labels.go
+++ b/engine/labels/labels.go
@@ -22,12 +22,18 @@ type LabelSelector[T Labeled] struct {
 	language gval.Language
 }
 
+<<<<<<< HEAD
 //Helper function to refactor key value label manipulation
 func AsString(key, value string) string {
     if value == "" {
         return fmt.Sprintf("%s", key)
     }
     return fmt.Sprintf("%s=%s", key, value)
+=======
+// Helper function to refactor key value label manipulation
+func formatLabel(key, value string) string{
+	return fmt.Sprintf("%s=%s", key, value)
+>>>>>>> 5d4e9ee20df94d06832dd0fbf058987f81105bbe
 }
 
 func (l *LabelSelector[T]) Matches(v T) (bool, error) {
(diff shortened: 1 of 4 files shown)


engine/labels/labels.go
```diff
@@ -22,6 +22,11 @@ type LabelSelector[T Labeled] struct {
 	language gval.Language
 }
 
+// Helper function to refactor key value label manipulation
+func formatLabel(key, value string) string{
```
pranavgaikwad, line 26: Can we export this function so that we don't need all the other functions elsewhere? And maybe call it `AsString(key, value string)` or something. If value is empty, we should be removing the trailing "=" in the label.

external-providers/golang-external-provider/pkg/golang/dependency.go
```diff
@@ -16,6 +16,17 @@ const (
 	golangDownloadableDepSourceLabel = "downloadable"
 )
 
+// createDepSourceLabel creates a dependency source label with the specified value.
+func createDepSourceLabel(label, value string) string {
```
pranavgaikwad, line 20: These two functions do the exact same thing. We should use the  exposed function from labels package instead.

provider/internal/java/dependency.go
```diff
@@ -26,6 +26,11 @@ const (
 	providerSpecificConfigExcludePackagesKey   = "excludePackages"
 )
 
+// createDepLabelItem creates a label item with the provided key-value pair.
+func createDepLabelItem(key, value string) string {
```
pranavgaikwad, line 30: same as above...use the exported function

provider/internal/java/dependency_test.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/go-logr/logr/testr"
 	"github.com/konveyor/analyzer-lsp/provider"
 )
+// createDepLabelItem creates a label item with the provided key-value pair.
+func createLabelItem(key, value string) string {
```
pranavgaikwad, line 13: no need to define another function in the test, should be able to use exported function from labels package

pranavgaikwad (changes requested):

RipulHandoo:
I encountered an error while raising the Pull Request in this branch. The error message was [insert error message here]. As a result, I'll be creating a new Pull Request to address the issue. If anyone has encountered a similar problem or has any insights, your help would be appreciated.

RipulHandoo:
I encountered an error while raising the Pull Request in this branch
. As a result, I'll be creating a new Pull Request to address the issue.

RipulHandoo:
I encountered an error while raising the Pull Request in this branch.  As a result, I'll be creating a new Pull Request to address the issue.

PR #266, opened 2023-07-19
:sparkles: Generic Provider

Chanakya-TS:
A generic provider that can be used to create an external provider for any language that is compliant with LSP 3.17 specifications.

Adds `go` provider that is initialized using the generic provider binary.


demo-output.yaml
```diff
@@ -166,26 +166,6 @@
         variables:
           file: file:///analyzer-lsp/examples/golang/main.go
       extras: []
-    golang-gomod-dependencies:
```
fabianvf, line 169: Don't commit changes to this file, we don't want the output to be different so if this file is changed it means the changeover isn't complete

external-providers/golang-dependency-provider/main.go
```diff
@@ -17,42 +19,47 @@ const (
...
-		return nil, nil
+		log.Fatal(fmt.Errorf("no dependencies"))
+		return
 	}
 
-	m := map[uri.URI][]*provider.Dep{}
+	m := map[uri.URI][]provider.Dep{}
 	for u, d := range ll {
-		m[u] = provider.ConvertDagItemsToList(d)
+		m[u] = ConvertDagItemsToList(d)
```
fabianvf, line 35: why not use the implementation in `provider`?

external-providers/golang-dependency-provider/main.go
```diff
@@ -17,42 +19,47 @@ const (
...
-func (g *golangServiceClient) GetDependencies() (map[uri.URI][]*provider.Dep, error) {
-	ll, err := g.GetDependenciesDAG()
-	if err != nil {
-		return nil, err
+		log.Fatal(err)
+		return
 	}
 	if len(ll) == 0 {
-		return nil, nil
+		log.Fatal(fmt.Errorf("no dependencies"))
```
fabianvf, line 29: I don't think we want to error out, sometimes there may be no dependencies

external-providers/generic-external-provider/pkg/generic/dependency.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (g *genericServiceClient) GetDependencies() (map[uri.URI][]*provider.Dep, error) {
+	buf := bytes.Buffer{}
+	cmdStr, isString := g.config.ProviderSpecificConfig["dependencyProviderPath"].(string)
+	if !isString {
+		return nil, fmt.Errorf("dependency provider path is not a string")
+	}
+	cmd := exec.Command(cmdStr)
+	cmd.Dir = g.config.Location
+	cmd.Stdout = &buf
```
fabianvf, line 22: I think you'd typically capture output more like this:

 ```
 out, err := cmd.Output()
```

rather than overwriting the stdout buffer

**Review by fabianvf — CHANGES_REQUESTED** [review]

File: demo-output.yaml (hunk 1 above)
Comment (Chanakya-TS) [reply demo-output.yaml]: my bad, the file was included accidentally while i was commiting

File: external-providers/golang-dependency-provider/main.go (hunk 2 above)
Comment (Chanakya-TS) [reply external-providers/golang-dependency-provider/main.go:35]: i thought having a `*provider.Dep` would interfere with marshalling, I tried it out with the implementation in `provider` and it works, so ill change that.

File: external-providers/golang-dependency-provider/main.go (hunk 3 above)
Comment (Chanakya-TS) [reply external-providers/golang-dependency-provider/main.go:29]: that was left out from when i was trying to debug, i was checking if i was getting any output from the binary

File: external-providers/generic-external-provider/pkg/generic/dependency.go (hunk 4 above)
Comment (Chanakya-TS) [reply external-providers/generic-external-provider/pkg/generic/dependency.go]: aah yes, ill do that

File: external-providers/generic-external-provider/pkg/generic/dependency.go [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return nil, err
+	}
+	data := buf.String()
+	if len(data) == 0 {
+		log.Fatal("no data")
+		return nil, fmt.Errorf("nO DEPENDENCIES")
+	}
+	m := map[uri.URI][]provider.Dep{}
+	json.Unmarshal([]byte(data), &m)
```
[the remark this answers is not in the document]
Chanakya-TS: thank you! that fixed it!

external-providers/generic-external-provider/pkg/generic/dependency.go
Chanakya-TS: @fabianvf should i also do the same when I take the output from `go mod graph` ?

docs/providers.md
```diff
@@ -29,24 +29,35 @@ If an explicit `proxyConfig` is not specified for a provider, system-wide proxy
...
+                "name": "go",
+                "lspServerPath": "/path/to/language/server/binary",
+                "dependencyProviderPath": "/path/to/dependency/provider/binary"
+            }
         }
     ]
 }
 ```
 
+The `generic provider` takes the following options in `providerSpecificConfig`:
```
fabianvf, line 56: We probably also need to add a field to `providerSpecificConfig` that allow us to specify arguments to pass to the language server, as not all of them will be able to just run with no arguments like the golang one

external-providers/generic-external-provider/go.mod
```diff
@@ -38,5 +38,3 @@ require (
 	google.golang.org/protobuf v1.30.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
-
-replace github.com/konveyor/analyzer-lsp => ../../
```
fabianvf, line 42: I think we want to leave this, right?

external-providers/generic-external-provider/pkg/generic/provider.go
```diff
@@ -57,7 +57,7 @@ func (p *golangProvider) Init(ctx context.Context, log logr.Logger, c provider.I
 	}
 
 	ctx, cancelFunc := context.WithCancel(ctx)
-	log = log.WithValues("provider", "golang")
+	log = log.WithValues("provider", c.ProviderSpecificConfig["name"])
```
fabianvf, line 60: This option was not mentioned in the documentation

fabianvf (changes requested):
I think you'll also want to remove the now-unused golang provider code

external-providers/generic-external-provider/go.mod
Chanakya-TS: I think that was automatically removed, when i ran `go mod tidy`. I will add it back.

external-providers/generic-external-provider/pkg/generic/provider.go
Chanakya-TS: Thank you, will add it to the documentation.

demo.Dockerfile
```diff
@@ -4,4 +4,5 @@ WORKDIR /analyzer-lsp
 
 COPY rule-example.yaml /analyzer-lsp/rule-example.yaml
 COPY examples /analyzer-lsp/examples
+COPY examples/golang /analyzer-lsp/external-providers/golang-dependency-provider/golang-dependency-provider
```
fabianvf, line 7: Is this intentional?  I think we should leave the examples under `/analyzer-lsp/examples`

docs/providers.md
```diff
@@ -29,24 +29,40 @@ If an explicit `proxyConfig` is not specified for a provider, system-wide proxy
 ```Note For Java: full analysis mode will search all the dependency and source, source-only will only search the source code. for a Jar/Ear/War, this is the code that is compiled in that archive and nothing else.
 ```
 
-#### Go provider
+#### Generic provider
+y
```
fabianvf, line 33: Looks like an extra character snuck in here

external-providers/golang-dependency-provider/main.go
```diff
@@ -18,42 +20,45 @@ const (
...
 		m[u] = provider.ConvertDagItemsToList(d)
 	}
 
-	return m, err
+	jsonStr, err := json.Marshal(m)
+	if err != nil {
+		log.Fatal(fmt.Errorf("unable to marshal dependencies"))
+		return
+	}
+	fmt.Println(string(jsonStr))
```
fabianvf, line 43: Should use the `log`

external-providers/generic-external-provider/pkg/generic/provider.go
```diff
@@ -37,11 +37,11 @@ func (p *golangProvider) Capabilities() []provider.Capability {
 	}
 }
 
-type golangCondition struct {
+type genericCondition struct {
 	Referenced string `yaml:"referenced"`
```
fabianvf, line 41: We should actually make `Referenced` a struct here, and give it a field called `Pattern` instead of having `Referenced` be a string itself. This allows us to expand the default implementation of `Referenced` in the future without making breaking changes to the structure of rules (ie, if we added  a generic `Location` field that allowed you to differentiate between declarations and usages).

fabianvf (changes requested):

docs/experimental/python-provider.md
```diff
@@ -0,0 +1,11 @@
...
+
+We are using the jedi-language-server (https://github.com/pappasam/jedi-language-server) to make a python provider using the generic provider.
+
+jedi-language-server can be installed using
+
+```
+pip install jedi-langauge-server
+```
+
+It will be installed in `/home/<user_name>/.local/bin/jedi-language-server`
```
fabianvf, line 11: you could also add the example config/sample main.py/example rule just for the sake of completeness

fabianvf (approved):

external-providers/generic-external-provider/pkg/generic/dependency.go
```diff
@@ -0,0 +1,35 @@
...
+	dataR, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+	data := string(dataR)
+	if len(data) == 0 {
+		return nil, nil
+	}
+	m := map[uri.URI][]*provider.Dep{}
+	json.Unmarshal([]byte(data), &m)
```
pranavgaikwad, line 29: Handle unmarshal error

pranavgaikwad (changes requested):
looks good overall, small concern

external-providers/generic-external-provider/pkg/generic/dependency.go
fabianvf: nice catch

external-providers/golang-dependency-provider/main.go
```diff
@@ -18,42 +20,47 @@ const (
...
+		return
 	}
 
 	m := map[uri.URI][]*provider.Dep{}
 	for u, d := range ll {
 		m[u] = provider.ConvertDagItemsToList(d)
 	}
 
-	return m, err
+	jsonStr, err := json.Marshal(m)
```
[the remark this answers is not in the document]
pranavgaikwad: okay, ignore above, just spoke with Fabian

pranavgaikwad (approved):

PR #270, opened 2023-07-25
:sparkles: Deduplicate dependencies from a single source

### LINKED ISSUE #259: Dependencies in the flat list should be deduplicated
The deduplication should consider Name, Version and SHA. There is one scenario to consider here that a dependency could be a direct and indirect both (coming via some other dep). In that case, we should remove the indirect dependency in the flat list.

fabianvf:
closes #259


fabianvf:
not able to replicate the test failure locally, trying to figure out why

pranavgaikwad (approved):

PR #271, opened 2023-07-27
:book: fix the Providers and Rules link in README.md file

Parthiba-Hazra:
providers and rules links are not working properly..


pranavgaikwad:
@Parthiba-Hazra thank you for fixing that!

Issue #175, opened 2023-05-24
Do not fail when dependency SHA cannot be read from file


pranavgaikwad:
Right now, we fail and return if any one dependency SHA cant be found. We need to continue looking at other deps.

shawn-hurley:
you can find this code specifically for the java provider here:
https://github.com/konveyor/analyzer-lsp/blob/main/provider/internal/java/dependency.go#L202

pranavgaikwad:
@punithnayak thank you! please take it away

punithnayak:
Hey @pranavgaikwad 
![image](https://github.com/konveyor/analyzer-lsp/assets/110713318/3d0aea16-35e2-4778-9db1-b42d5e14193a)
Facing this issue

punithnayak:
Hey @pranavgaikwad can you please tell  me how resolve this issue

shawn-hurley:
I think that you need to add docker.io as a registry

pranavgaikwad:
Fixed via #272

PR #272, opened 2023-07-27
:bug: Not to fail when dependency SHA cannot be read from file.

Parthiba-Hazra:
Hey in failure scenario I set `d.ResolvedIdentifier` as an empty string (d.ResolvedIdentifier = "").
Let me know if I need to set it to any other value.
I think this is the only function that need to update maybe have to update also `parseMavenDepLines` function .. let me know if I am wrong.

Issue: #175

### DIFF
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -258,9 +258,14 @@ func (p *javaServiceClient) parseDepString(dep, localRepoPath string) (provider.
 	fp := filepath.Join(localRepoPath, strings.Replace(parts[0], ".", "/", -1), parts[1], d.Version, fmt.Sprintf("%v-%v.jar.sha1", parts[1], d.Version))
 	b, err := os.ReadFile(fp)
 	if err != nil {
-		return d, err
+		// Log the error and continue with the next dependency.
+		fmt.Printf("Error reading SHA-1 hash file for dependency %s: %s\n", dep, err)
+		// Set some default or empty resolved identifier for the dependency.
+		d.ResolvedIdentifier = ""
+	} else {
+		d.ResolvedIdentifier = string(b)
 	}
-	d.ResolvedIdentifier = string(b)
+
 	d.Labels = p.addDepLabels(d.Name)
 	d.FileURIPrefix = fmt.Sprintf("%v://contents%v", FILE_URI_PREFIX, filepath.Dir(fp))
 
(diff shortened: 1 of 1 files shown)


provider/internal/java/dependency.go
```diff
@@ -258,9 +258,14 @@ func (p *javaServiceClient) parseDepString(dep, localRepoPath string) (provider.
 	fp := filepath.Join(localRepoPath, strings.Replace(parts[0], ".", "/", -1), parts[1], d.Version, fmt.Sprintf("%v-%v.jar.sha1", parts[1], d.Version))
 	b, err := os.ReadFile(fp)
 	if err != nil {
-		return d, err
+		// Log the error and continue with the next dependency.
+		fmt.Printf("Error reading SHA hash file for dependency %s: %s\n", dep, err)
```
pranavgaikwad, line 262: can you please change the `fmt.Printf` with logger instead? See example usage of the provider logger here https://github.com/konveyor/analyzer-lsp/blob/728d78ba9b9ff5dd2cc4b7acda2aed8a58b612b7/provider/internal/java/dependency.go#L348

That uses info function, we would use the Error() function of the logger instead

pranavgaikwad (commented):
Thank you for your contribution! Requesting a small change

provider/internal/java/dependency.go
Parthiba-Hazra: yeah sure ..

provider/internal/java/dependency.go
```diff
@@ -258,9 +258,14 @@ func (p *javaServiceClient) parseDepString(dep, localRepoPath string) (provider.
 	fp := filepath.Join(localRepoPath, strings.Replace(parts[0], ".", "/", -1), parts[1], d.Version, fmt.Sprintf("%v-%v.jar.sha1", parts[1], d.Version))
 	b, err := os.ReadFile(fp)
 	if err != nil {
-		return d, err
+		// Log the error and continue with the next dependency.
+		p.log.V(7).Info("Error reading SHA hash file for dependency %s: %s\n", dep, err)
```
pranavgaikwad, line 262:
```suggestion
		p.log.V(5).Error(err, "error reading SHA hash file for dependency", "dep", d.Name)
```

pranavgaikwad (changes requested):

provider/internal/java/dependency.go
Parthiba-Hazra: oh my bad .. thanks for the suggestion. will update now

pranavgaikwad (approved):

Issue #273, opened 2023-07-27
Add validation on provider names


pranavgaikwad:
Since we are now integrating external providers, we should have validations on provider names such that there are no duplicates. We should fail right away before initializing any of the providers if we find duplicates in provider config.

Parthiba-Hazra:
Hey @pranavgaikwad 
I submitted a pr on this can you please check out..

pranavgaikwad:
Fixed via #275

PR #275, opened 2023-07-28
:bug: Fail right away before initializing any of the providers if we find duplicates in provider config

Parthiba-Hazra:
provider configs should not contain duplicate providers. 

issue: #273

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -103,6 +103,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Validate provider names for duplicate providers.
+	if err := validateProviderName(configs); err != nil {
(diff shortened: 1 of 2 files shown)


pranavgaikwad:
@Parthiba-Hazra I think itd be better if we handled validation in GetConfig(). That way anyone using that function will get an error on ambiguous provider names in the config.

Parthiba-Hazra:
> @Parthiba-Hazra I think itd be better if we handled validation in provider.GetConfig(). That way anyone using that function will get an error on ambiguous provider names in the config.

Ok no problem I will update that.

shawn-hurley (approved):
Looks good to me

pranavgaikwad (approved):

PR #276, opened 2023-07-31
:bug: do not use dependency file URIs in conditions

pranavgaikwad:


pranavgaikwad:
@shawn-hurley When there are dependency URIs in one of chained conditions, engine panics on the Filename() call:

```
 \t/usr/local/go/src/runtime/panic.go:838 +0x207"
 go.lsp.dev/uri.URI.Filename(...)'
 \t/go/pkg/mod/go.lsp.dev/uri@v0.3.0/uri.go:52"
 github.com/konveyor/analyzer-lsp/engine.incidentsToFilepaths({0xc0004b1680?, 0x10, 0xc000e1ddd0?})'
 \t/analyzer-lsp/engine/conditions.go:268 +0x1a5"
 github.com/konveyor/analyzer-lsp/engine.OrCondition.Evaluate({{0xc00024c380?, 0x4?, 0x4?}}, {0xc8e378?, 0xc000e1dda0?}, {{0xc8f198?, 0xc0004d4f80?}, 0x0?}, {0xc000b9db60, 0xc000e1dd70})'
 \t/analyzer-lsp/engine/conditions.go:225 +0x42a"
```

Note that this doesn't actually filter out from violations, this is only for chained conditions where filepaths are passed as input to descendents. I agree though, instead of taking them out, maybe we pass the File URIs as-is without doing Filename() call.

shawn-hurley:
The filename will remove the `file://` prefix IIUC and return that file path. Removing Filename(), call is probably correct if the URI does not have the file scheme.

pranavgaikwad:
Updated

fabianvf (approved):

Issue #278, opened 2023-08-01
Do not generate violations when there are no incidents


pranavgaikwad:
We filter out incidents based on dep label selector. When we filter all incidents out, we shouldn't be generating violation as it doesn't mean anything.

Parthiba-Hazra:
> @Parthiba-Hazra that sounds correct, thank you! It'd be nice to log how many incidents we filtered though.

yea go it, will create a pr soon on this ..

PR #281, opened 2023-08-02
Do not generate violations when there are no incidents

Parthiba-Hazra:
issue: #278

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -401,12 +401,14 @@ func (p *ProviderCondition) Evaluate(ctx context.Context, log logr.Logger, condC
 	}
 
 	incidents := []engine.IncidentContext{}
+	filteredOutCount := 0 // Counter to track incidents filtered out
 	for _, inc := range resp.Incidents {
 		// filter out incidents that don't match the dep label selector
 		if matched, err := matchDepLabelSelector(p.DepLabelSelector, inc, deps); err != nil {
 			log.V(5).Error(err, "failed to match dep label selector")
 			return engine.ConditionResponse{}, err
 		} else if !matched {
+			filteredOutCount++
 			continue
 		}
 		i := engine.IncidentContext{
@@ -431,6 +433,13 @@ func (p *ProviderCondition) Evaluate(ctx context.Context, log logr.Logger, condC
 		}
 		incidents = append(incidents, i)
 	}
+
+	// If there are no incidents, don't generate any violations
+	if len(incidents) == 0 {
+		log.V(3).Info("filtered out all incidents based on dep label selector", "filteredOutCount: ", filteredOutCount)
+		return engine.ConditionResponse{}, nil
+	}
+
 	cr := engine.ConditionResponse{
 		Matched:         resp.Matched,
 		TemplateContext: resp.TemplateContext,
@@ -438,6 +447,7 @@ func (p *ProviderCondition) Evaluate(ctx context.Context, log logr.Logger, condC
 	}
 
 	log.V(8).Info("condition response", "ruleID", p.Rule.RuleID, "response", cr, "cap", p.Capability, "conditionInfo", p.ConditionInfo, "client", p.Client)
+	log.V(3).Info("filtered out incidents based on dep label selector", "filteredOutCount: ", filteredOutCount)
 	return cr, nil
 
 }


Parthiba-Hazra:
@pranavgaikwad the v-level in logger instance is right?

provider/provider.go
```diff
@@ -431,13 +433,21 @@ func (p *ProviderCondition) Evaluate(ctx context.Context, log logr.Logger, condC
 		}
 		incidents = append(incidents, i)
 	}
+
+	// If there are no incidents, don't generate any violations
+	if len(incidents) == 0 {
```
pranavgaikwad, line 438: note that there's nothing stopping a provider from returning 0 incidents, we need to be certain that this happens only when we filtered something. `(len(incidents) == 0 && filteredCount > 0)`. Also extreme nit :).... no need to introduce a new variable, just use diff `len(resp.incidents) - len(incidents)` to get filteredCount

provider/provider.go
```diff
@@ -431,13 +433,21 @@ func (p *ProviderCondition) Evaluate(ctx context.Context, log logr.Logger, condC
...
+	}
+
 	cr := engine.ConditionResponse{
 		Matched:         resp.Matched,
 		TemplateContext: resp.TemplateContext,
 		Incidents:       incidents,
 	}
 
 	log.V(8).Info("condition response", "ruleID", p.Rule.RuleID, "response", cr, "cap", p.Capability, "conditionInfo", p.ConditionInfo, "client", p.Client)
+	log.V(3).Info("filtered out incidents based on dep label selector", "filteredOutCount: ", filteredOutCount)
```
pranavgaikwad, line 450: only log when filtered

provider/provider.go
```diff
@@ -431,13 +433,21 @@ func (p *ProviderCondition) Evaluate(ctx context.Context, log logr.Logger, condC
 		}
 		incidents = append(incidents, i)
 	}
+
+	// If there are no incidents, don't generate any violations
+	if len(incidents) == 0 {
+		log.V(3).Info("filtered out all incidents based on dep label selector", "filteredOutCount: ", filteredOutCount)
```
pranavgaikwad, line 439:
```suggestion
		log.V(5).Info("filtered out all incidents based on dep label selector", "filteredOutCount", filteredOutCount)
```

pranavgaikwad (changes requested):

provider/provider.go
Parthiba-Hazra: yea.. that's right 
I miss the point

pranavgaikwad:
@Parthiba-Hazra Hey, I just came across another issue that is related to the changes you made here, are you willing to include that fix here? 

The issue is that we don't want to apply label selector when evaluating tagging rules, that is because, tagging rules do not generate incidents, they only care about whether a dependency condition was matched or not. That would mean that when we filter out all incidents, we still want to return the value of original `resp.Matched`, but not return incidents. The filtering will happen in the engine instead of the provider condition. Happy to explain this in more detail if you want.

Parthiba-Hazra:
> @Parthiba-Hazra Hey, I just came across another issue that is related to the changes you made here, are you willing to include that fix here?
> 
> The issue is that we don't want to apply label selector when evaluating tagging rules, that is because, tagging rules do not generate incidents, they only care about whether a dependency condition was matched or not. That would mean that when we filter out all incidents, we still want to return the value of original `resp.Matched`, but not return incidents. The filtering will happen in the engine instead of the provider condition. Happy to explain this in more detail if you want.

Hi @pranavgaikwad sorry for delay response.
I would love to fix that also. Let me check out it first.
I will reach out to you in slack for further clarification, thank you.

pranavgaikwad:
@shawn-hurley This PR makes it so that violations are not created when there are no incidents. This was an issue because we were seeing violations being created even when all incidents are filtered out. It doesn't really mean anything. 

I wonder if there's any other case I might be missing where we want violations even when there are no incidents.

provider/provider.go
```diff
@@ -431,13 +431,25 @@ func (p *ProviderCondition) Evaluate(ctx context.Context, log logr.Logger, condC
 		}
 		incidents = append(incidents, i)
 	}
+
+	// If there are no incidents, don't generate any violations
+	if len(incidents) == 0 && len(resp.Incidents)-len(incidents) > 0 {
+		log.V(5).Info("filtered out all incidents based on dep label selector", "filteredOutCount", len(resp.Incidents)-len(incidents))
```
shawn-hurley, line 437: When is v(5) logged again? it is not in the default logs just if you increase verbosity right?

shawn-hurley (approved):
I agree, we do not want to have violations if we can not point to why we have violated. I believe this to be true fro info rules as well so I am a +1 on this change.

I do want to be careful with the logs here to make sure they are part of verbose logging and not default but not blocking if you disagree @pranavgaikwad

provider/provider.go
shawn-hurley: I think it should be fine, don't worry about it. I think that in the future, we need guidelines for verbosity levels, but because we don't have any, this probably means it is a followup. Resolving the conversation and this LGTM

PR #285, opened 2023-08-04
:sparkles: Refactor multiline strings to improve human-readability

RipulHandoo:
I've addressed the issue by manually refactoring the multiline code snippets to enhance their human-readability. The gopkg.in/yaml.v2 library was not automatically formatting the code snippets as multiline strings, resulting in less readable output. I've made the necessary adjustments so that the code snippets are now presented in a more organized and understandable format.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -421,15 +421,7 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 			if err != nil || codeSnip == "" {
 				r.logger.V(6).Error(err, "unable to get code location")
 			} else {
-				// Split the code snippet into lines
-				lines := strings.Split(codeSnip, "\n")
-
-				// Build the properly formatted multiline string
-				var formattedCodeSnip string
-				for _, line := range lines {
-					formattedCodeSnip += "    " + line + "\n"
-				}
-				incident.CodeSnip = "|-\n" + formattedCodeSnip
+				incident.CodeSnip = fmt.Sprintf("%s\n", codeSnip)
 			}
 			fileCodeSnipCount[string(m.FileURI)] += 1
 		}
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -421,7 +421,15 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 			if err != nil || codeSnip == "" {
 				r.logger.V(6).Error(err, "unable to get code location")
 			} else {
-				incident.CodeSnip = codeSnip
+				// Split the code snippet into lines
+				lines := strings.Split(codeSnip, "\n")
+
+				// Build the properly formatted multiline string
+				var formattedCodeSnip string
+				for _, line := range lines {
+					formattedCodeSnip += "    " + line + "\n"
+				}
+				incident.CodeSnip = "|-\n" + formattedCodeSnip
 			}
 			fileCodeSnipCount[string(m.FileURI)] += 1
 		}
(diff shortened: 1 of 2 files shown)


pranavgaikwad:
@RipulHandoo We need to make the code change that ensures multiline snippets are generated as expected, I think all you need to do is append a newline to codesnip...I could be wrong. But worth a try

RipulHandoo:
> @RipulHandoo We need to make the code change that ensures multiline snippets are generated as expected, I think all you need to do is append a newline to codesnip...I could be wrong. But worth a try

@pranavgaikwad i have append a newline to codesnip like this
``` 
codeSnip: | 
//remaining lines
```
But at the time of testing this is the error last 3 lines
``` 
\ No newline at end of file
+  - lang-ref-002
Error: Process completed with exit code 1.
```

pranavgaikwad:
@RipulHandoo manually updating the demo-output.yaml file is not going to help, it's a file that's used as expected output in the CI to compare the results that come from an actual analysis. Since we didn't make any code change, analyzer is always going to produce the output with the "ugly" code snips and fail the CI because expected output has "pretty" output. To fix this issue, we not only need to change the demo-output.yaml file, but also make a code change that ensures analyzer always outputs the pretty code snips.

RipulHandoo:
> @RipulHandoo manually updating the demo-output.yaml file is not going to help, it's a file that's used as expected output in the CI to compare the results that come from an actual analysis. Since we didn't make any code change, analyzer is always going to produce the output with the "ugly" code snips and fail the CI because expected output has "pretty" output. To fix this issue, we not only need to change the demo-output.yaml file, but also make a code change that ensures analyzer always outputs the pretty code snips.

@pranavgaikwad Where can i find that code?
And can you Recommend what particular changes are required?

pranavgaikwad:
@RipulHandoo https://github.com/konveyor/analyzer-lsp/blob/846335e641b2cbdb7b4755f2a783d1f151ba6580/engine/engine.go#L420-L424

RipulHandoo:
@pranavgaikwad 
Does this good to you
```
codeSnip, err := r.getCodeLocation(m, rule)
			if err != nil || codeSnip == "" {
				r.logger.V(6).Error(err, "unable to get code location")
			} else {
				formattedLines := []string{"codeSnip: |"}

				lines := strings.Split(codeSnip, "\n")

				for i, line := range lines {
					formattedLine := fmt.Sprintf("          %2d  %s", i+1, line)
					formattedLines = append(formattedLines, formattedLine)
				}

				formattedCodeSnip := strings.Join(formattedLines, "\n")

				incident.CodeSnip = formattedCodeSnip
			}
```

engine/engine.go
```diff
@@ -421,7 +421,15 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 			if err != nil || codeSnip == "" {
 				r.logger.V(6).Error(err, "unable to get code location")
 			} else {
-				incident.CodeSnip = codeSnip
+				// Split the code snippet into lines
```
pranavgaikwad, line 424: I think all you need to do here is `incident.CodeSnip = fmt.Sprintf("%s\n", codeSnip)` If that doesn't fix it, probably newlines are not going to fix the issue

pranavgaikwad:
I discussed this with @RipulHandoo and our initial assumption about adding newlines doesnt seem to solve the issue. I will need some time to investigate it better, closing this for now as we have decided not to pursue this solution any further.

PR #286, opened 2023-08-07
:sparkles: use default maven index, stop providers before exiting in dependency cli

### LINKED ISSUE #300: Suppress dep output to stdout
We already write the output to a file, we should replace output to stdout with a log line that says something like  "generated output to <>". I ask this because dependency output can be huge and it pollutes the logs someone might actually be interested in seeing.

pranavgaikwad:
Closes #300 

- Uses default maven index in the image to classify open-source deps
- Improves dependency CLI a little by stopping the providers and removing unnecessy output to stdout
- Disables Jaeger by default, as it requires users to run it via entrypoint command when running outside CI test


pranavgaikwad:
this is dependent on java-analyzer-bundle build, it's failing right now due to disk space / timing issues.

pranavgaikwad:
this should work, there is only one internal dependency in this app, but in demo output I see some others labeled internal too. investigating

shawn-hurley (approved):
/lgtm

PR #288, opened 2023-08-09
:sparkles: Adds a flag context-lines to set CONTEXT_LINES variable in engine.go

### LINKED ISSUE #282: Make CONTEXT_LINES for code snips configureable
This will allow users to tweak the value based on how they see fit, we will still keep the default number low to keep memory utilization low.

MdSahil-oss:
This PR Fixes #282


MdSahil-oss:
@pranavgaikwad Please review this PR :)

cmd/analyzer/main.go
```diff
@@ -103,6 +104,7 @@ func main() {
 		os.Exit(1)
 	}
 
+	engine.CONTEXT_LINES = *contextLines
```
fabianvf, line 107: I think rather than changing a global variable, we'll want to plumb the value through the structs, similarly to `incidentLimit`. We may also need to plumb this through the providers as they can implement custom codesnippers as well.

MdSahil-oss: @fabianvf done 👍🏽

pranavgaikwad (changes requested):
@MdSahil-oss we also need to pass this setting to the providers in providerSpecificConfig. Right now, only the Java provider gives its own code snippets here https://github.com/konveyor/analyzer-lsp/blob/main/provider/internal/java/snipper.go

MdSahil-oss:
@pranavgaikwad Passed contextLines to the providers in a diffrent way than your suggetion, please review this PR again :)

pranavgaikwad:
@MdSahil-oss Thank you! I don't mind that change @fabianvf would you take another look?

fabianvf (approved):
Nice, I like that approach much better actually

shawn-hurley (approved):
Love the use of the options!

pranavgaikwad (approved):

PR #292, opened 2023-08-11
:bug: only decompile for code snips when file isnt present

pranavgaikwad:
It looks like we are doing unnecessary work multiple times for the same dep in a given run decompiling everytime, we should just do it once


shawn-hurley (approved):
I like this +1

Issue #241, opened 2023-06-30
Our CodeSnips are ugly


fabianvf:
The `gopkg.in/yaml.v2` library is not turning our codesnips into `|` style multiline strings, it's just leaving them as string literals with `\n` in there. Ideally we'd output prettier multiline strings because in the current state it's not really human-readable at all.

fabianvf:
The issue is as far as I can tell the library doesn't currently have any way to enforce that, if there are certain (undocumented AFAIK) characters in the string it won't use the multi-line syntax

MdSahil-oss:
@fabianvf @pranavgaikwad I would like to work on this issue.

pranavgaikwad:
@MdSahil-oss In the `rule-example.yaml` file, `lang-ref-001` rule is a good reproducer for this one https://github.com/konveyor/analyzer-lsp/blob/7ebd2f5efb3fbfbb5e102d80e8481ee823fdb3b2/demo-output.yaml#L210

RipulHandoo:
Hey @fabianvf @pranavgaikwad , I have manually modified the [demo-output.yaml](https://github.com/RipulHandoo/analyzer-lsp/blob/main/demo-output.yamlrl)
And pulled a    [PR](https://github.com/konveyor/analyzer-lsp/pull/285) but getting error .
How to overcome that, ```yaml``` file is valid I double-checked that

RipulHandoo:
> Hey @fabianvf @pranavgaikwad , I have manually modified the [demo-output.yaml](https://github.com/RipulHandoo/analyzer-lsp/blob/main/demo-output.yamlrl) And pulled a [PR](https://github.com/konveyor/analyzer-lsp/pull/285) but getting error . How to overcome that, `yaml` file is valid I double-checked that

In this
 ```
codeSnip: |
          1  package com.example.apps;
          2  
          3  import io.fabric8.kubernetes.api.model.apiextensions.v1beta1.CustomResourceDefinition;
          4  
          5  public class App 
          6  {
          7  
          8      public static void main( String[] args )
          9      {
          10          CustomResourceDefinition crd = new CustomResourceDefinition();
          11          System.out.println( crd );
          12  
          13          GenericClass<String> element = new GenericClass<String>(\"Hello world!\");
          14          element.get();
          15      }
          16  }
```
Do the line number are also have to remove line number and only code lines needs to be in the ```CodeSnips:```

Obasoro:
[...]

PR #293, opened 2023-08-15
:sparkles: Our CodeSnips are ugly

AdiAkhileshSingh15:
This pull request addresses issue #241 .
It aims to enhance the readability of codeSnip script in output YAML. After going through the codebase, I found that as we are treating the `CodeSnip` as a string. Even while marshaling the rulesets using the `gopkg.in/yaml.v2` package, the `CodeSnip` remained a string literal, regardless of multiline or single-line string.

To improve this, I wrote the outputViolations file with the new yaml content that has the `CodeSnip` as an object that encapsulates the code script, by storing the whole block of codeSnip object along with codeSnip object declaration in a new variable as 
`x := "codeSnip: |\n" + incidents[j].CodeSnip` 
`codesnip = append(codesnip, x)` 
The new yaml file was constructed by iteratively inserting codeSnip lines after each message object. But, here as you might have noticed that it will only work if for each message object there is a codeSnip object as well.

I have tried this in a sample code with the same yaml file demo-output.yaml, in that, after each message object it did insert the codeSnip as expected. We just need to add some \t  escape sequences before the block of codeSnip.

Although this approach may seem unconventional, it was born out of exhaustive attempts to maintain the desired code output format. Despite various trials, the existing limitations of marshaling to YAML with the `CodeSnip` as a string persisted.

While the code may need minor adjustments, I'm ready to make those changes. However, I kindly ask for your approval before proceeding with this approach of creating a temporary file and replacing it with our outputViolations file.

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


pranavgaikwad:
@AdiAkhileshSingh15 Thank you for you contribution! While your approach solves the issue, I don't think this is the best approach. I would rather fix the issue at the root where it occurs, rather than writing a wrapper that updates an existing output.yaml file.

pranavgaikwad:
Closing this in favor of #294

PR #294, opened 2023-08-16
:sparkles: Making CodeSnips Pretty

AdiAkhileshSingh15:
This pull request addresses issue #241 . It aims to improve the readability of the `codeSnip` script in the output YAML. I had previously submitted a pull request (#293 ) for the same issue, but that approach was quite lengthy and unconventional. Upon further investigation into why the `codesnip` wasn't being marshaled into the desired '|' block scalar style literal, I discovered that any `\s` (whitespace) character following a `\n` (newline) caused the string to be treated as a regular string. To resolve this, I removed any spaces that appeared after the `\n` in the `incident.CodeSnip`. 
@pranavgaikwad Sir, can you please check this.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -462,6 +462,8 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 			}
 		}
 
+		makeCodeSnipsBlockConvertible(&incident.CodeSnip)
+
 		if rule.Perform.Message.Text != nil {
 			variables := make(map[string]interface{})
 			for key, value := range m.Variables {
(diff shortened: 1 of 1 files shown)


pranavgaikwad:
@AdiAkhileshSingh15 Thank you for your PR! While this solves the issue, it appears that this is manipulating the original snip which we don't want because code snips are also displayed in the UI. I feel that there may not be a sane solution to this issue.

AdiAkhileshSingh15:
@pranavgaikwad. While this approach does indeed manipulate the original codesnip, it's important to note that it could result in an ugly UI only if maintaining the tab alignment at the start of each line is important, and is not maintained otherwise in the UI.

pranavgaikwad:
@AdiAkhileshSingh15 UI displays code snippets in a code editor, if we take out spaces / tabs from original snip, that will mess up the alignment of lines

pranavgaikwad:
@AdiAkhileshSingh15 apologies, I should have researched this a little more before commenting on the issue. I think my comment is leading folks to believe default yaml marshaling is just going to work. Looks like it won't. Closing this PR...and updating issue. Please feel free to re-open or create new PR if you do solve this correctly.

AdiAkhileshSingh15:
> @AdiAkhileshSingh15 apologies, I should have researched this a little more before commenting on the issue. I think my comment is leading folks to believe default yaml marshaling is just going to work. Looks like it won't. Closing this PR...and updating issue. Please feel free to re-open or create new PR if you do solve this correctly.

Ohkay! No problem.
And, yeah sure, will update you if i find a correct solution for this issue.

Issue #295, opened 2023-08-17
Deduplicate labels in violations


pranavgaikwad:
When creating violations, we should deduplicate labels inherited from rules. This happens because we copy labels from ruleset too, I think regardless of what we do elsewhere, labels should be deduplicated on violations.

AdiAkhileshSingh15:
I would like to work on this issue. To confirm my understanding: it seems we need to deduplicate labels in violations. These labels were copied from rule, but as the ruleset already contains them, these violation labels aren't utilized anywhere. If i'm correct, I have a plan to move forward with resolving this. Additionally, if you could assist me in reproducing the issue, your guidance would be highly appreciated.

pranavgaikwad:
@AdiAkhileshSingh15 Please take it away. Even though labels arent used in the codebase anywhere, they are important and we want to keep them on violations. To reproduce the issue, you can pass a `--rules` option pointing to a directory of rules, in there, you will need a `ruleset.yaml` file that will have the same labels as the rules. But, you can short circuit all this by just passing a rule that contains duplicate labels.

AdiAkhileshSingh15:
```
// labels on ruleset apply to all rules in it
			rule.Labels = append(rule.Labels, ruleSet.Labels...)
```
So @pranavgaikwad ,if am not wrong, is it that when we assign labels from a ruleset to specific rules, there might be instances of duplicate labels? Consequently, when we pass these labels from rule, in violations, should we there implement a process to filter out these duplicates?

pranavgaikwad:
@AdiAkhileshSingh15 yeah, i think just deduplicate labels at the time of adding to violations in createViolation() function

pranavgaikwad:
Fixed via #296

PR #296, opened 2023-08-17
:sparkles: Deduplicating Labels

AdiAkhileshSingh15:
This pull request addresses issue #295.  I resolved the problem, of duplicate labels appearing both in violations and rules after adding labels from the ruleset, by introducing a function that flags all duplicate values within the labels array.
@pranavgaikwad please check.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -491,6 +491,8 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 		}
 	}
 
+	deduplicateLabels(&rule.Labels)
+
 	return konveyor.Violation{
 		Description: rule.Description,
 		Labels:      rule.Labels,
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -557,3 +559,18 @@ func matchesAllSelectors(m RuleMeta, selectors ...RuleSelector) bool {
 	}
 	return true
 }
+
+func deduplicateLabels(labels *[]string) {
```
pranavgaikwad, line 563: @AdiAkhileshSingh15 Thank you for taking this! can we not use pointers here, and just return a new array?

pranavgaikwad (changes requested):

engine/engine.go
AdiAkhileshSingh15: > @AdiAkhileshSingh15 Thank you for taking this! can we not use pointers here, and just return a new array?

We surely can. Initially, I leaned towards pointers for efficiency, but as returning a new array of unique labels can make the code more readable, we can surely do it that way.

pranavgaikwad (approved):

Issue #236, opened 2023-06-28
Add jaeger to image so it runs out of the box


pranavgaikwad:
There are two changes needed - one in demo docker, and another one in base image. In the demo docker file, we need to do two things...first, run jaeger prior to running the analyzer...second, after analyzer finishes, make a request to jaeger to download the traces as json. These changes will go in the `demo.Dockerfile`. In the base image, we only need to include jaeger, i.e. the first step.

Parthiba-Hazra:
HI @pranavgaikwad I placed a pr on this ..please checkout.

PR #297, opened 2023-08-18
:sparkles: Added jaeger to images so it runs out of the box

Parthiba-Hazra:
Fixes: #236

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,8 +16,19 @@ COPY  Makefile /analyzer-lsp/Makefile
 
 RUN make build
 
+FROM debian:buster AS jaeger-builder
+WORKDIR /jaeger
+
+RUN apt-get update && \
+    apt-get install -y curl && \
+    curl -L -o jaeger-1.47.0-linux-amd64.tar.gz https://github.com/jaegertracing/jaeger/releases/download/v1.47.0/jaeger-1.47.0-linux-amd64.tar.gz && \
+    tar -xzf jaeger-1.47.0-linux-amd64.tar.gz && \
+    rm jaeger-1.47.0-linux-amd64.tar.gz
+
 # The unofficial base image w/ jdtls and gopls installed
 FROM quay.io/konveyor/jdtls-server-base
+
+COPY --from=jaeger-builder /jaeger/jaeger-1.47.0-linux-amd64/* /usr/bin/
 
 COPY --from=builder /analyzer-lsp/konveyor-analyzer /usr/bin/konveyor-analyzer
 COPY --from=builder /analyzer-lsp/konveyor-analyzer-dep /usr/bin/konveyor-analyzer-dep
@@ -28,4 +39,6 @@ COPY provider_container_settings.json /analyzer-lsp/provider_settings.json
 
 WORKDIR /analyzer-lsp
 
-ENTRYPOINT ["konveyor-analyzer"]
+EXPOSE 5775/udp 6831/udp 6832/udp 5778 16686 14268 9411
+
+ENTRYPOINT ["sh", "-c", "jaeger-all-in-one && konveyor-analyzer"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,22 +16,12 @@ COPY  Makefile /analyzer-lsp/Makefile
 
 RUN make build
 
-FROM debian:buster AS jaeger-builder
-WORKDIR /jaeger
-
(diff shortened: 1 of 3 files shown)


pranavgaikwad:
@Parthiba-Hazra Thank you for your PR! Can we not build Jaeger and directly include the [jaeger/all-in-one](https://hub.docker.com/r/jaegertracing/all-in-one) image? Also, can we remove duplication in demo.Dockerfile and use the binary directly from base layer?

Parthiba-Hazra:
@pranavgaikwad please review the changes.. sorry for the trouble actually I had to use command `all-in-one-linux` instead of `jaeger-all-in-one`. :)

demo.Dockerfile
```diff
@@ -1,7 +1,15 @@
+FROM jaegertracing/all-in-one:latest AS jaeger-builder
```
pranavgaikwad, line 1: I don't think we need this line here, the binary should already be available in the image `quay.io/konveyor/analyzer-lsp`

demo.Dockerfile
```diff
@@ -1,7 +1,15 @@
...
+
 FROM quay.io/konveyor/analyzer-lsp
 
+COPY --from=jaeger-builder /go/bin/all-in-one-linux /usr/bin/
```
pranavgaikwad, line 5: Same as above, not needed

pranavgaikwad (changes requested):

demo.Dockerfile
```diff
@@ -1,7 +1,15 @@
...
+
 WORKDIR /analyzer-lsp
 
 COPY rule-example.yaml /analyzer-lsp/rule-example.yaml
 COPY examples /analyzer-lsp/examples
 COPY open-source-libs.txt /analyzer-lsp/open-source-libs.txt
+
+EXPOSE 5775/udp 6831/udp 6832/udp 5778 16686 14268 9411
+
+ENTRYPOINT ["sh", "-c", "all-in-one-linux & sleep 5 && konveyor-analyzer --enable-jaeger=true --jaeger-endpoint=http://localhost:14268/api/traces && curl -o traces.json http://localhost:16686/api/traces?service=analyzer-lsp"]
```
pranavgaikwad, line 15: this `--jaeger-endpoint` is already the default, no need to explicitely pass https://github.com/konveyor/analyzer-lsp/blob/7cb1d667a4088a4ba11711491ff98987ce80c3ab/cmd/analyzer/main.go#L58

pranavgaikwad (changes requested):
Can we also enable jaeger by default? instead of having users to pass it, now that it's already in the image, it should just work

demo.Dockerfile
Parthiba-Hazra: --enable-jaeger=true already there.. I think I just need to remove the --jaeger-endpoint flag?

pranavgaikwad: @Parthiba-Hazra I meant in the analysis CLI here https://github.com/konveyor/analyzer-lsp/blob/7cb1d667a4088a4ba11711491ff98987ce80c3ab/cmd/analyzer/main.go#L57

Parthiba-Hazra: Done.. thanks for the guidance :)

Dockerfile
```diff
@@ -28,4 +32,6 @@ COPY provider_container_settings.json /analyzer-lsp/provider_settings.json
 
 WORKDIR /analyzer-lsp
 
-ENTRYPOINT ["konveyor-analyzer"]
+EXPOSE 5775/udp 6831/udp 6832/udp 5778 16686 14268 9411
```
pranavgaikwad, line 35: @Parthiba-Hazra thank you for incorporating the changes. One last thing, I missed this earlier, sorry. Can we only expose the 16686 port, as all other ports are only used internally?

pranavgaikwad (changes requested):

Dockerfile
Parthiba-Hazra: done :+1:

pranavgaikwad (approved):

pranavgaikwad:
@Parthiba-Hazra thank you! appreciate your patience on this. Going to to give this a quick test before merging.

pranavgaikwad:
@shawn-hurley any objection on using jaeger image from docker.io?

shawn-hurley:
None from me, I am a plus one

Issue #235, opened 2023-06-28
Consider refactoring code that deals with condition response and violations


pranavgaikwad:
It seems that theres some repetation in the areas where we deal with condition responses and where we create violations. Consider refactoring that part.

MdSahil-oss:
@pranavgaikwad I would like to work on it.

AdiAkhileshSingh15:
@pranavgaikwad I went through the code that deals with condition responses and violations, mainly the part where we create violations. In the process, I identified sections of code that remain unused due to repitition or overlapping fields of different objects. For instance, the "links" field within the Violation object. Could you provide further details on other areas where you've noticed repetition, suggesting the need for refactoring?

PR #298, opened 2023-08-20
:sparkles: Refactors code that deals with condition response and violations

AdiAkhileshSingh15:
This pull request addresses issue #235 .
Removes unused code that was redundant as it's functionality was handled by other elements. Actually, the `links` field in `IncidentContext` draws it's value from `rule`, and as `rule` is also passed to `createViolations` function, we can derive the `links` in the function itself. So, the `links` field within `IncidentContext` is currently unused throughout the source code; its primary role was to transfer this value to `violations`, which is handled by `rule`. With your approval, we can safely remove this field to enhance code clarity.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -405,15 +405,6 @@ func (r *ruleEngine) createViolation(conditionResponse ConditionResponse, rule R
 			lineNumber := *m.LineNumber
 			incident.LineNumber = &lineNumber
 		}
-		links := []konveyor.Link{}
-		if len(m.Links) > 0 {
-			for _, l := range m.Links {
-				links = append(links, konveyor.Link{
-					URL:   l.URL,
-					Title: l.Title,
-				})
-			}
-		}
 		// Some violations may not have a location in code.
 		limitSnip := (r.codeSnipLimit != 0 && fileCodeSnipCount[string(m.FileURI)] == r.codeSnipLimit)
 		if !limitSnip {


pranavgaikwad:
@AdiAkhileshSingh15 thank you for looking through code to investigate. This is definitely something to fix, but the refactoring the original issue is talking about spans wider than this fix. @shawn-hurley brought that issue up and I think the expectation was to reduce the complexity around parts of the engine / providers in general.

shawn-hurley:
I would probably suggest that we make #235 a tracking issue and create a new sub issue to fix this, and merge it IMO 

Thoughts @pranavgaikwad @AdiAkhileshSingh15

AdiAkhileshSingh15:
> @AdiAkhileshSingh15 thank you for looking through code to investigate. This is definitely something to fix, but the refactoring the original issue is talking about spans wider than this fix. @shawn-hurley brought that issue up and I think the expectation was to reduce the complexity around parts of the engine / providers in general.

@pranavgaikwad 
Ohkay! I've been trying to figure out the right spots for refactoring within the codebase. However, due to the interconnections and complexity of the objects, I've found it a bit challenging to pinpoint the exact areas that require refactoring. If you and @shawn-hurley could provide some guidance on this, I'd be absolutely delighted to contribute to this. As mentioned by @shawn-hurley, we could turn this into a tracking issue and then break it down into sub-issues that offer more detail, specifically addressing what parts truly need refactoring.

PR #304, opened 2023-08-22
:sparkles: add fallbacks for when workspace/symbol isn't giving us what we need

fabianvf:
- ~~This is currently failing with Go for... some reason. It can't load packages anymore?~~
- ~~This is probably going to be hideously slow on large projects, the directory tree search will need to be parallelized before this is a feasible approach~~
- I haven't verified that this works with other providers. We should give the experimental python one another try

### DIFF
--- a/external-providers/generic-external-provider/main.go
+++ b/external-providers/generic-external-provider/main.go
@@ -35,4 +35,22 @@ func main() {
 	s := provider.NewServer(client, *port, log)
 	ctx := context.TODO()
 	s.Start(ctx)
+	// serviceClient, err := client.Init(ctx, log, provider.InitConfig{
+	// 	Location:     "/home/fabian/projects/github.com/konveyor/analyzer-lsp/examples/golang",
+	// 	AnalysisMode: "full",
+	// 	ProviderSpecificConfig: map[string]interface{}{
+	// 		"name":          "go",
+	// 		"lspServerPath": "gopls",
+	// 		"lspArgs":       []interface{}{"-vv", "-logfile", "go-debug.log", "-rpc.trace"},
+	// 	},
+	// })
+	// if err != nil {
+	// 	panic(err)
+	// }
+
+	// response, err := serviceClient.Evaluate("referenced", []byte(`{"referenced":{pattern: ".*v1beta1.CustomResourceDefinition"}}`))
+	// if err != nil {
+	// 	panic(err)
+	// }
+	// fmt.Println(response)
 }
--- a/external-providers/generic-external-provider/pkg/generic/service_client.go
+++ b/external-providers/generic-external-provider/pkg/generic/service_client.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"context"
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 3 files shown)


external-providers/generic-external-provider/main.go
```diff
@@ -35,4 +35,22 @@ func main() {
 	s := provider.NewServer(client, *port, log)
 	ctx := context.TODO()
 	s.Start(ctx)
+	// serviceClient, err := client.Init(ctx, log, provider.InitConfig{
```
fabianvf, line 38: We should remove this, but being able to run the provider directly has been really nice for debugging. Maybe should add a debug subcommand or something for this.

external-providers/generic-external-provider/pkg/generic/service_client.go
```diff
@@ -75,29 +79,119 @@ func (p *genericServiceClient) Evaluate(cap string, conditionInfo []byte) (provi
 }
 
 func (p *genericServiceClient) GetAllSymbols(query string) []protocol.WorkspaceSymbol {
+	logger, err := os.OpenFile("golang-provider.log",
```
fabianvf, line 82: Should be removed, this was just for debugging when running from analyzer-lsp

external-providers/generic-external-provider/main.go
shawn-hurley: make it a hidden subcommand in the main binary?

fabianvf: I don't think it can work in the main binary, would have to be in the provider binaries since the communication between the two is over RPC. Here you can delve debug / see the print statements.

Side note, we really should be capturing the stdout from these providers and presenting them to the users, I don't think we do currently (right?)

shawn-hurley: Not in a real way, There is the plumbing to do it though

We have one set of questions with that, though,

1. Do we print it out as it happens, or do we bunch them together once a given evaluation is finished
2. Are there log levels that need to be returned, is this a verbosity thing where we only print on high levels?

Let me know if you want me to take a stab at it. @pranavgaikwad I am out of the loop on what is important now.

rule-example.yaml
```diff
@@ -43,12 +43,12 @@
...
         location: TYPE
     - go.referenced: 
-        pattern: "v1beta1.CustomResourceDefinition"
+        pattern: ".*v1beta1.CustomResourceDefinition"
 - message: 'golang apiextensions/v1/customresourcedefinitions found {{file}}:{{lineNumber}}'
   ruleID: go-lang-ref-001
   when:
     go.referenced: 
-        pattern: "v1beta1.CustomResourceDefinition"
+        pattern: ".*v1beta1.CustomResourceDefinition"
```
shawn-hurley, line 51: why is this needed?

fabianvf: This should make it trigger the fallback

external-providers/generic-external-provider/main.go
fabianvf: I think we should print them as they happen, I wouldn't necessarily bother with verbosity levels as a first-class thing since a provider can always implement verbosity levels through `providerSpecificConfig`. So we'd just want to log with an identifier that lets you know which provider name is logging and do it in real time so that hangs/etc can be debugged

fabianvf: I think if you wanted to take a stab at it it would be a huge boon to developing providers. such a PITA right now

pranavgaikwad: @shawn-hurley @fabianvf let me know if this sums up the work https://github.com/konveyor/analyzer-lsp/issues/312

shawn-hurley: will do then!

shawn-hurley (approved):

pranavgaikwad (approved):

Issue #305, opened 2023-08-24
Analyze command should not run analysis when --help option is set


pranavgaikwad:
It looks like we run the analyzer even when someone uses --help option, we should exit instead.

PR #307, opened 2023-08-28
:bug: analyze command should not run analysis when --help option is set

Parthiba-Hazra:
closes: #305


pranavgaikwad:
@Parthiba-Hazra thank you!

PR #309, opened 2023-08-29
:bug: unify conditional interface implementations to either use value or pointer types

### LINKED ISSUE #301: Unify Conditional interface implementations to either use value or pointer types
I was looking at the implementations of the Conditional interface the other day and saw that of the 5 types that implement it (AndCondition, ConditionEntry, OrCondition, ProviderCondition, DependencyCondition), only ProviderCondition implements the interface as a pointer:

https://github.com/konveyor/analyzer-lsp/blob/7cb1d667a4088a4ba11711491ff98987ce80c3ab/engine/conditions.go#L73-L75

https://github.com/konveyor/analyzer-lsp/blob/7cb1d667a4088a4ba11711491ff98987ce80c3ab/provider/provider.go#L359

The other types implement the interface as a value type:

https://github.com/konveyor/analyzer-lsp/blob/7cb1d667a4088a4ba11711491ff98987ce80c3ab/engine/conditions.go#L197

I was wondering if we should unify this so we either work with value types or pointers. Looking to get your thoughts @pranavgaikwad @shawn-hurley

JonahSussman:
Fixes #301

### DIFF
--- a/parser/rule_parser.go
+++ b/parser/rule_parser.go
@@ -778,7 +778,7 @@ func (r *RuleParser) getConditionForProvider(langProvider, capability string, va
 		selector = r.DepLabelSelector
 	}
 
-	return &provider.ProviderCondition{
+	return provider.ProviderCondition{
 		Client:           client,
 		Capability:       capability,
 		ConditionInfo:    value,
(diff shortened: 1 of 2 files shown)


shawn-hurley:
@JonahSussman This looks correct,  you just have to add the Signoff and re-push

pranavgaikwad (approved):

PR #310, opened 2023-08-30
:warning: always use up-to-date context

pranavgaikwad:
We were using stale context. I observed it when looking at tracer output, some spans were being recorded in the scope of main context, instead of their provider specific context. 

:warning: changes provider api

### DIFF
--- a/provider/internal/java/util.go
+++ b/provider/internal/java/util.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	"github.com/konveyor/analyzer-lsp/tracing"
 )
 
 const javaProjectPom = `<?xml version="1.0" encoding="UTF-8"?>
@@ -41,6 +42,9 @@ const javaProjectPom = `<?xml version="1.0" encoding="UTF-8"?>
 // creates new java project and puts the java files in the tree of the project
 // returns path to exploded archive, path to java project, and an error when encountered
 func decompileJava(ctx context.Context, log logr.Logger, archivePath string) (explodedPath, projectPath string, err error) {
+	ctx, span := tracing.StartNewSpan(ctx, "decompile")
+	defer span.End()
+
 	projectPath = filepath.Join(filepath.Dir(archivePath), "java-project")
 
 	err = createJavaProject(ctx, projectPath)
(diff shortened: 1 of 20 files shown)


provider/internal/java/util.go
```diff
@@ -41,6 +42,9 @@ const javaProjectPom = `<?xml version="1.0" encoding="UTF-8"?>
 // creates new java project and puts the java files in the tree of the project
 // returns path to exploded archive, path to java project, and an error when encountered
 func decompileJava(ctx context.Context, log logr.Logger, archivePath string) (explodedPath, projectPath string, err error) {
+	ctx, span := tracing.StartNewSpan(ctx, "decompile")
```
djzager, line 45: Nice.

djzager (approved):
VISACK

shawn-hurley (commented):
We don't want the ctx that the service clients have for initialization to be the same as the ctx for a specific call to Evaluate or Dep or do we?

shawn-hurley:
Could there be a time when the engine is run multiple times? I think that is the use case?

pranavgaikwad:
@shawn-hurley I don't think I completely understand, as long as we pass the same context to engine and all providers here https://github.com/konveyor/analyzer-lsp/blob/6ad9f5e29e94ab2c8080c690cf56c5610bb7de08/cmd/analyzer/main.go#L164 and here https://github.com/konveyor/analyzer-lsp/blob/6ad9f5e29e94ab2c8080c690cf56c5610bb7de08/cmd/analyzer/main.go#L136. shouldn't that be fine?

shawn-hurley:
Yeah, but could there be a time when the engine is running on its own, and rules come in, and the providers are instantiated when the rules come in (if not set)? Stoping one call to then engine then is different than stopping the full engine.

You are not wrong that right now, this will work exactly as you expect, I think keeping them separate prevents some hard debugging in future IMO

If we don't think that this is ever a concern that it is not a huge deal.

shawn-hurley:
makes sense, then I am good with this change

shawn-hurley:
@pranavgaikwad can you rebase?

pranavgaikwad:
@shawn-hurley rebased.

shawn-hurley:
Sorry need another rebase, I am sorry I tried to reduce this

PR #330, opened 2023-09-12
:sparkles: Adding ability to snip files and make all konveyor-jdt into file URI's

shawn-hurley:


provider/internal/java/snipper.go
```diff
@@ -20,66 +18,15 @@ const (
...
-			}
-		}
-
-		snip, err := p.scanFile(javaFileAbsolutePath, loc)
-		if err != nil {
-			fmt.Printf("\n%v", err)
-			return "", err
-		}
-		return snip, nil
+		fmt.Printf("\n%v", err)
```
shawn-hurley, line 26: I need to clean up, but I need to test with the internal dep

provider/internal/java/filter.go
```diff
@@ -191,10 +183,61 @@ func (p *javaServiceClient) convertSymbolRefToIncidentContext(symbol protocol.Wo
...
+	packageName := u.Query().Get("packageName")
+
+	var jarPath string
+	if sourceRange {
+		// If there is a source range, we know we know there is a sources jar
+		jarName := filepath.Base(u.Path)
+		s := strings.TrimSuffix(jarName, ".jar")
+		s = fmt.Sprintf("%v-sources.jar", s)
+		jarPath = filepath.Join(filepath.Dir(u.Path), s)
+	} else {
```
pranavgaikwad, line 218: interesting, so we did find the solution for non source?

pranavgaikwad (approved):

provider/internal/java/filter.go
shawn-hurley: I don't know if this is working as intended and/or fixes the full problem. It does solve the other big problem that folks are bringing up, so it is worth getting in, and then I will add some follow up after testing tomorrow to see how off it all is.

PR #331, opened 2023-09-13
Add omitempty for Violation Extras

eemcmullan:
For https://github.com/konveyor-ecosystem/kantra we need to allow an option to generate the reports in JSON. The JSON marshal was failing as this field was leaving an empty map.


pranavgaikwad:
@eemcmullan please remove extras fields from the demo-output.yaml and this should be good to go

pranavgaikwad (changes requested):
Update demo-output.yaml to remove `extras: []` from violations

pranavgaikwad (approved):

PR #342, opened 2023-09-26
:bug: Update the label value format  as per new format

pranavgaikwad:
Forgot to update the regex which actually does the match, also added a new unit test to catch that flow in future

### DIFF
--- a/engine/labels/labels_test.go
+++ b/engine/labels/labels_test.go
@@ -159,6 +159,12 @@ func TestParseLabel(t *testing.T) {
 			wantKey: "konveyor.io/fact",
 			wantVal: "Spring Beans",
 		},
+		{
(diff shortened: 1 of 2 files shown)


engine/labels/labels_test.go
```diff
@@ -159,6 +159,12 @@ func TestParseLabel(t *testing.T) {
 			wantKey: "konveyor.io/fact",
 			wantVal: "Spring Beans",
 		},
+		{
+			name:    "version ranges in values",
```
fabianvf, line 163: should we add tests for the `+`, `-` variations as well?

pranavgaikwad: @fabianvf added

fabianvf (approved):

PR #345, opened 2023-09-28
:bug: fix wildcard values in labels

### LINKED ISSUE #344: [BUG] Rules with konveyor.io/source or konveyor.io/target wildcard match labels should match on any target

### Konveyor version

latest

### Priority

Major

### Current Behavior

The rules with these wildcard labels are skipped

### Expected Behavior

The rules should be matched

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Create a rule with label `konveyor.io/source`
2. Pass `--label-selector=konveyor.io/source=eap` as input when running analysis.

### Environment

```markdown
- **Analyzer Image SHA** latest
- **JDTLS  base image SHA** latest
- **Provider Settings JSON** latest
```

### Anything else?

_No response_

pranavgaikwad:
Closes #344

### DIFF
--- a/engine/labels/labels_test.go
+++ b/engine/labels/labels_test.go
@@ -360,6 +360,23 @@ func Test_ruleSelector_Matches(t *testing.T) {
 			},
 			want: true,
 		},
+		{
+			name: "simple wildcard value",
+			expr: "(konveyor.io/source=technology-usage)",
+			ruleLabels: []string{
+				"konveyor.io/source",
+			},
+			want: true,
+		},
+		{
+			name: "wildcard value expression",
+			expr: "(konveyor.io/target && konveyor.io/source=test)",
+			ruleLabels: []string{
+				"konveyor.io/target=eap",
+				"konveyor.io/source",
+			},
+			want: true,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/engine/labels/labels.go
+++ b/engine/labels/labels.go
@@ -251,6 +251,10 @@ func matchesAny(elem string, items []string) bool {
 // returns true when names of values are equal and the version of
 // candidate falls within the version range of matchWith
 func labelValueMatches(matchWith string, candidate string) bool {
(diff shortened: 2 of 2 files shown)


engine/labels/labels_test.go
```diff
@@ -360,6 +360,23 @@ func Test_ruleSelector_Matches(t *testing.T) {
...
+			name: "simple wildcard value",
+			expr: "(konveyor.io/source=technology-usage)",
+			ruleLabels: []string{
+				"konveyor.io/source",
+			},
+			want: true,
+		},
+		{
+			name: "wildcard value expression",
+			expr: "(konveyor.io/target && konveyor.io/source=test)",
```
fabianvf, line 373: Hmm, this reads a little confusing to me

This looks like it says, `konveyor.io/target` exists at all, and the `konveyor.io/source` has to exist, and be equal to `test`

But then the label below just has a `source`, with no value, which I guess we're basically treating like `source=*`. Does this mean there's no way to exclude a ruleset that doesn't provide a value for a label?  I guess we'd have that same behavior whether we treated `*` or `""` as the wildcard value so maybe it doesn't matter. 

Feel like we'll need a lot of docs around this behavior though because I find it pretty unintuitive just looking at the selector vs labels

fabianvf (approved):
Think it's probably too late in the process to change it up, but maybe warrants some further discussion, or at least ensuring that we've got enough docs to cover it.

engine/labels/labels_test.go
pranavgaikwad: @fabianvf to exclude, you will need to do a `!konveyor.io/source`  or to specifically exclude something with `test` value `!konveyor.io/source=target`

fabianvf: yeah it makes sense, just think we'll need to doc it well. It's more intuitive to have wildcards in the selectors than the labels, that's probably what's throwing me off.

pranavgaikwad:
Closing this as we have decided to make this a more explicit by adding a label on rules we want to run all the time

PR #347, opened 2023-09-28
:sparkles: Add a label to short-circuit selector

pranavgaikwad:


docs/labels.md
```diff
@@ -29,14 +29,15 @@ labels:
...
+## Rule Labels
 
-The analyzer defines some labels that have special meaning. Here is a list of all such labels:
+The analyzer defines some labels that have special meanings:
 
-- `konveyor.io/source`: Identifies source technology a rule or a ruleset applies to.
-- `konveyor.io/target`: Identifies target technology a rule or a ruleset applies to.
+- `konveyor.io/source`: Identifies source technology a rule or a ruleset applies to. The value can be a string with optional version range at the end e.g. "eap", "eap6", "eap7-" etc.
+- `konveyor.io/target`: Identifies target technology a rule or a ruleset applies to. The value can be a string with optional version range at the end e.g. "eap", "eap6", "eap8+" etc.
+- `konveyor.io/selector`: Defines rule disposition with respect to the [rule label selector](#rule-label-selector). A value of `always` indicates that the rule will always be included irrespective of the label selector, while a value of `never` indicates that the rule will always be excluded.
```
fabianvf, line 38: Hmm, is it confusing to re-use selector here since we already use that terminology for the CLI input?

fabianvf: just seems strange because I can't imagine it having meaning past `always` and `never`, but from the name I feel like I should be able to toss any selector string in there (though I've no idea what it could mean)

djzager (approved):
ACK

PR #348, opened 2023-09-28
:sparkles: Add preliminary python support

JonahSussman:
Big PR, hopefully I didn't break anything


.github/workflows/demo-testing.yml
```diff
@@ -14,15 +14,23 @@ jobs:
...
 
+      - name: install yq for testing
+        run: go install github.com/mikefarah/yq/v4@latest
+
       - name: run demo image and ensure violations output unchanged
         run: |
           podman run -v $(pwd)/demo-output.yaml:/analyzer-lsp/output.yaml:Z localhost/testing:latest
-          git diff --exit-code demo-output.yaml
+          diff \
+            <(sort <(sed 's/incidents\.[0-9]\+/incidents\.x/g' <(yq -P 'sort_keys(..)' -o=props <(git show HEAD:demo-output.yaml)))) \
```
pranavgaikwad, line 24: what sorcery is this?

JonahSussman: Since the incidents in a violation can be returned in any order, `git diff` kept failing. i.e. `git diff` would compare

```yaml
sample-rule:
  incidents:
  - A
  - B
```
```yaml
sample-rule:
  incidents:
  - B
  - A
```

and fail This is because diff only compares the _textual differences_ between the files. It has no knowledge of the YAML structure. To fix this, I installed [`yq`](https://mikefarah.gitbook.io/yq) which allows us to examine the actual structure the input YAML. My whole goal was to still use `diff` to compare the two files, but make the position of each incident in the array irrelevant.

Moving from right to left, I first get the different versions of `demo-output.yaml`. I then pipe that into `yq`, with the [`-o=props` flag ](https://mikefarah.gitbook.io/yq/usage/properties).  `-P 'sort_keys(..)'` is actually unnecessary due to what I do later. If we piped the YAML files I showed into that, it would return:

```yaml
sample-rule.incidents.0 = A
sample-rule.incidents.1 = B
```
```yaml
sample-rule.incidents.0 = B
sample-rule.incidents.1 = A
```

I then pipe that output into `sed`, replacing every `incidents.<number>` with `incidents.x` to erase which index each element is in. Finally, I sort the result and `diff` the two streams. Both files would return

```yaml
sample-rule.incidents.x = A
sample-rule.incidents.x = B
```

Which is the same thing. No error, ta-da!

There might be a better way than cobbling together some shell commands for this. I wouldn't even be surprised if there was a way you could do all of this in `yq`! At least one of the advantages of having it installed is that we can investigate the actual structure of the YAML now in our tests.

pranavgaikwad: @JonahSussman lovely, I think this is a much better version of what we are doing in Kantra now, would love to copy it there

JonahSussman: I would consider it a great honor, copy away! I would remove the `-P 'sort_keys(..)'` part as I'm almost positive it's redundant.

external-providers/generic-external-provider/main.go
```diff
@@ -10,13 +10,109 @@ import (
...
+	// need to do research on mapping in logrusr to level here TODO
+	logrusLog.SetLevel(logrus.Level(5))
+
+	log := logrusr.New(logrusLog)
+
+	ctx := context.TODO()
+	client := generic.NewGenericProvider()
+
+	golangSC, err := client.Init(ctx, log, provider.InitConfig{
+		Location: "/home/jonah/Projects/analyzer-lsp/examples/golang",
```
djzager, line 41: This path doesn't look like one we'd want to commit.

external-providers/generic-external-provider/pkg/generic/service_client.go
```diff
@@ -144,85 +180,144 @@ func parallelWalk(location string, regex *regexp.Regexp) ([]protocol.TextDocumen
 	return positions, nil
 }
 
-func (p *genericServiceClient) GetAllSymbols(ctx context.Context, query string) []protocol.WorkspaceSymbol {
+// Returns all symbols for the given query.
+// NOTE: Only returns definitions when server does not supoprt workspace/symbol.
+// Is is intended behavior?
```
djzager, line 185: For the .NET work, what made the most sense while experimenting was to:
1. find occurrences of pattern
2. call "textDocument/definition" and see if the response URI includes the referenced namespace (ie. System.Web.Mvc)...if it does then we add it as an incident.

And, iirc @fabianvf had mentioned gopls handling `references` differently which would mean that this is yet another area where differences in implementation may bite us. Makes me wonder if we want to make it possible to pass a `callback` of sorts that allows  you to specify a function for determining if a found Symbol using the fallback method should be included as an incident.

lsp/protocol/extensions.go
```diff
@@ -0,0 +1,151 @@
+package protocol
```
djzager, line 1: I am hopeful that we can move to https://github.com/go-language-server/protocol but I think that's an `important-long-term` type issue. Do we agree? I can writeup the issue later.

.github/workflows/demo-testing.yml
shawn-hurley: Can we sort these in the engine then?

external-providers/generic-external-provider/main.go
shawn-hurley: IMO, Eiher we remove this, or we make this more generic for testing. 

I would think that removing for this PR and then coming back and making it easy to test w/ the generic provider makes sense to me. OR we make this a feature of kantra, and work on it from that the layer? 

IDK just don't know if we want to change this particular main file yet but I could be wrong

provider/provider.go
```diff
@@ -102,12 +102,25 @@ const (
 type InitConfig struct {
 	// This is the location of the code base that the
 	// Provider will be responisble for parsing
+	// Deprecated: rootUri, which is what this maps to in the LSP spec, is deprecated.
```
shawn-hurley, line 105: Can you create issues in the repo for this and link in the deprecated or a TODO?

external-providers/generic-external-provider/main.go
JonahSussman: Whoops, will fix! Is leaving in the `mainDebug()` function a problem as well? I just found all the stuff I put in there extremely useful when I was having problems.

external-providers/generic-external-provider/pkg/generic/service_client.go
JonahSussman: If I'm understanding right, you're saying the .NET work utilizes the fallback method? I added the note as if we use the fallback method, only the definitions will be returned, rather than every symbol like when using `workspace/symbol`. Honestly, I'm not quite seeing why we should do anything other than the fallback method?

lsp/protocol/extensions.go
JonahSussman: Definitely something I want to investigate. Any way we can make it less brittle is a win in my book

JonahSussman:
> This looks good, a few comments but I don't believe any of them are blockers.
> 
> I am curious though, after looking into our fallback symbol searching, if we could accept some `func`s for the handling of some of the differences between different language servers. That may enable us to make the generic provider work down to the lowest common denominator and have specific examples for golang/pythong/typescript/etc on how to extend it to handle peculiarities between the different implementations.

@djzager I really want to investigate a good way to do this. The vast majority of the functionality between different providers seems the same. Would love to have a discussion about what that looks like

external-providers/generic-external-provider/pkg/generic/service_client.go
djzager: I copy/pasted and then modified the fallback method for what I was doing with the .NET work. See [here](https://github.com/konveyor/analyzer-dotnet-provider/pull/7/files#diff-908ae641af05346d94599b805eeb64ded6e475d6f461d8a06e09facdb4443291R81).

My thought was that it may be possible for us to create an exported method for our fallback that accepts a `func` for determining if a given symbol is the one the provider is looking for.

djzager: Looking more thoroughly at what I did, it may be best for us to have our fallback method as an exported function that simply returns a slice of positions...leaving it up to the caller on how to sort which of those positions are valid symbols or not.

external-providers/generic-external-provider/pkg/generic/service_client.go
```diff
@@ -19,21 +20,23 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-type genericServiceClient struct {
-	rpc        *jsonrpc2.Conn
-	cancelFunc context.CancelFunc
-	cmd        *exec.Cmd
+type GenericServiceClient struct {
```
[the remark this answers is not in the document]
JonahSussman: Had it for debugging stuff, changing it back

external-providers/generic-external-provider/pkg/generic/tables.go
```diff
@@ -0,0 +1,29 @@
+package generic
+
+import "strings"
+
+// Ideally we wouldn't have to do anything like this. Alas, we must.
+
+// 2-tiered table for each language and their respective fixes we may need to
+// apply
+var SubstitutionTable = map[string]map[string]string{
```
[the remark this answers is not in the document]
JonahSussman: Will add to the `python-provider.md` doc!

provider/provider.go
JonahSussman: Done! https://github.com/konveyor/analyzer-lsp/issues/365

external-providers/generic-external-provider/pkg/generic/provider.go
```diff
@@ -109,10 +109,10 @@ func (p *genericProvider) Init(ctx context.Context, log logr.Logger, c provider.
 	}()
 
 	svcClient := genericServiceClient{
-		rpc:        rpc,
-		cancelFunc: cancelFunc,
-		cmd:        cmd,
-		config:     c,
+		Rpc:        rpc,
```
shawn-hurley, line 112: nit: If changing, then it should be RPC

external-providers/generic-external-provider/pkg/generic/provider.go
```diff
@@ -109,10 +109,10 @@ func (p *genericProvider) Init(ctx context.Context, log logr.Logger, c provider.
...
+		CancelFunc: cancelFunc,
+		Cmd:        cmd,
```
shawn-hurley, line 114: `CMD` as well

jsonrpc2/jsonrpc2.go
```diff
@@ -127,6 +139,9 @@ func (c *Conn) Call(ctx context.Context, method string, params, result interface
 			h.Done(ctx, err)
 		}
 	}()
+
+	// fmt.Printf("Sending: %s\n", string(data))
```
shawn-hurley, line 143: lets remove this instead of commenting IMO

shawn-hurley (approved):
nit's but the core is good

PR #351, opened 2023-10-04
Don't use full file paths for issues

### LINKED ISSUE #322: Full file paths for dependencies are irrelevant for the user
### Environment

- Fedora 36
- Analysis configuration: 
  - Source + Dependencies
  - cloud-readiness, linux
  - No custom rules

### Description

Files in the affected files side drawer contain the full path to each file that contains incidents for the given issue. In the case of dependencies, this is displaying a long path to the location of the dependency JAR within the local .m2 repository available in the task pod. That information is completely irrelevant and potentially confusing for the user, that will only care about the JAR filename and the affected class/file within that JAR.

### How to reproduce

- Analyze tackle-test app with the configuration available in the Environment section.
- Go to the issues view, Single application, and filter by the tackle-testapp application.
- Select for example the _Java Mail API_ issue.
- Click on _View affected files_

### Expected result

JAR filenames alongside the path to the affected file within the JAR are displayed.

### Actual result

Full file paths are displayed for each file, not including the actual affected file inside the JAR as referenced in #321 

Screenshot from 2023-09-08 11-07-31

eemcmullan:
Closes #322


demo-output.yaml
```diff
@@ -271,15 +271,15 @@
...
+          file: /src/main/java/com/example/apps/Singleton.java
           kind: Class
           name: Singleton
       - uri: file:///analyzer-lsp/examples/java/src/main/java/com/example/apps/Singleton.java
         message: condition entries should evaluate out of order
         codeSnip: " 1  package com.example.apps;\n 2  \n 3  import javax.ejb.SessionBean;\n 4  import javax.ejb.Singleton;\n 5  \n 6  @Singleton\n 7  public class Bean implements SessionBean {\n 8  }\n"
         lineNumber: 7
         variables:
-          file: file:///analyzer-lsp/examples/java/src/main/java/com/example/apps/Singleton.java
+          file: /src/main/java/com/example/apps/Singleton.java
```
fabianvf, line 282: It looks like we're trimming off the `file://` prefix, which we shouldn't be

cmd/analyzer/main.go
```diff
@@ -195,7 +199,7 @@ func main() {
 		}
 	}
 
-	rulesets := eng.RunRules(ctx, ruleSets, selectors...)
+	rulesets := eng.RunRules(ctx, ruleSets, locations, selectors...)
```
shawn-hurley, line 202: Should we make this a variable on the Engine, something like location prefixes?

engine/engine.go
```diff
@@ -397,6 +398,19 @@ func (r *ruleEngine) createViolation(ctx context.Context, conditionResponse Cond
 		if r.incidentLimit != 0 && len(incidents) == r.incidentLimit {
 			break
 		}
+		var updatedFile string
+		for i := range locations {
+			tempFile := uri.File(locations[i])
+			if strings.Contains(string(m.FileURI), locations[i]) {
+				updatedFile = strings.TrimPrefix(string(m.FileURI), string(tempFile))
```
shawn-hurley, line 405: we should break out of the for loop here

engine/engine.go
```diff
@@ -397,6 +398,19 @@ func (r *ruleEngine) createViolation(ctx context.Context, conditionResponse Cond
...
+			}
+		}
+		for k := range m.Variables {
```
shawn-hurley, line 408: can you just do and then not loop through stuff?

```go
if _, ok := m.Vairables["file"]; ok {
    m.Variables["file"] = updatedFile
}
```

engine/engine.go
shawn-hurley: we can also add the file:// back here, and fix that issue

PR #352, opened 2023-10-04
:sparkles: Identify interesting Jars to decomp and improve decomp performance

### LINKED ISSUE #317: Improve decompilation to include interesting JARs
In order to improve decompile performance, we disabled decompiling every nested JAR (See #203 & #308 ) in an archive assuming that all the nested JARs are dependencies, and we probably don't want to analyze them...this assumption is true to some extent. While we don't want to decompile every other JAR in the archive, we do want to decompile some (for instance, an internal dependency). This requires us to add more logic in how we decompile things. First thought here is that we should probably not decompile known open-source JARs, but there isn't an easy way to figure out the GroupId / ArtifactId of a JAR just from its name.

### LINKED ISSUE #319: Issue not found in internal dependency
### Environment

- Fedora 36
- Analysis configuration: 
  - Source + Dependencies
  - cloud-readiness, linux
  - No custom rules

### Description

The _File system - Java IO_ issue present in the config-utils dependency from the tackle-testapp applications is not found during the analysis, even when it has been configured in source + dependencies mode and the issue is found in many other dependencies.

### How to reproduce

- Analyze tackle-testapp.
- Go to the _Issues_ view, Single application tab and filter by the tackle-testapp application.
- Click on the _File system - Java IO_ issue. 37 affected files are found.
Screenshot from 2023-09-08 10-55-04
- Browse the list of affected files in the side drawer. config-utils-1.0.0.jar is not there.

### Expected result

The incident is found in the ApplicationConfiguration class from the config-utils dependency, as it was found when running a Windup analysis:

Screenshot from 2023-09-08 10-58-30

### Actual result

There are no incidents for that issue found in config-utils:

Screenshot from 2023-09-08 10-57-48

pranavgaikwad:
Fixes #317 
Fixes #319 

Summary of changes:
* When decompiling binaries, for every JAR we find, we attempt to look at its metadata to get artifact and group. If we dont find metadata, we look it up on maven central using its sha. For all such JARs that we find accurate information about, we add them to java project's pom as dependencies. For all other JARs, we send them to decompile.

* In decompile, we are running concurrently. Also, instead of decompiling each class file individually, we are now decompiling whole JAR using fernflower and then exploding it (this is faster than individual .class file decompile)

* Prior to initing Java provider, we are now downloading sources for all dependencies. If we find any that don't have sources, we are passing them to decompile. 

* When getting dependencies for a binary, we are now using same logic from step 1 to get more fine grained info for a JAR.

### DIFF
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -78,12 +78,12 @@ func (p *javaServiceClient) GetDependencies(ctx context.Context) (map[uri.URI][]
 	return m, nil
 }
 
-func (p *javaServiceClient) getLocalRepoPath() string {
+func getMavenLocalRepoPath(mvnSettingsFile string) string {
 	args := []string{
 		"help:evaluate", "-Dexpression=settings.localRepository", "-q", "-DforceStdout",
 	}
-	if p.mvnSettingsFile != "" {
-		args = append(args, "-s", p.mvnSettingsFile)
+	if mvnSettingsFile != "" {
+		args = append(args, "-s", mvnSettingsFile)
 	}
 	cmd := exec.Command("mvn", args...)
 	var outb bytes.Buffer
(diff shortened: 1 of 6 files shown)


djzager (approved):

shawn-hurley (approved):
Looks good to me!

PR #366, opened 2023-10-09
:ghost: Ignore Jetbrains config folder (.idea)

jmle:
For using Goland


shawn-hurley:
Merging because CI failing is not related.

PR #368, opened 2023-10-09
:bug: Must compile for dependency:sources to work properly

jmle:
Obviously the downside to this is that this command will take considerably longer

### DIFF
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -114,7 +114,9 @@ func (p *javaServiceClient) GetDependenciesFallback(ctx context.Context, locatio
 		pomDeps = append(pomDeps, *pom.Dependencies...)
 	}
 	if pom.DependencyManagement != nil {
-		pomDeps = append(pomDeps, *pom.DependencyManagement.Dependencies...)
+		if pom.DependencyManagement.Dependencies != nil {
+			pomDeps = append(pomDeps, *pom.DependencyManagement.Dependencies...)
+		}
 	}
 
 	// add each dependency found
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -122,7 +122,7 @@ func (p *javaServiceClient) GetDependencyFallback(ctx context.Context) (map[uri.
 	}
 	deps := []*provider.Dep{}
 	dep := &provider.Dep{}
-	// TODO this is comedically janky
+	// TODO this is comically janky
 	for _, node := range list {
 		if node.Data == "groupId" {
 			if dep.Name != "" {
@@ -163,6 +163,10 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
 	moddir := filepath.Dir(path)
 
 	pom, err := gopom.Parse(path)
+	if err != nil {
+		return nil, err
+	}
+
 	args := []string{
 		"dependency:tree",
 		"-Djava.net.useSystemProxies=true",
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -18,6 +18,7 @@ import (
 	"github.com/konveyor/analyzer-lsp/engine/labels"
 	"github.com/konveyor/analyzer-lsp/output/v1/konveyor"
 	"github.com/konveyor/analyzer-lsp/provider"
+	"github.com/vifraa/gopom"
 	"go.lsp.dev/uri"
 )
 
@@ -160,11 +161,17 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
 	defer os.Remove(f.Name())
 
 	moddir := filepath.Dir(path)
+
+	pom, err := gopom.Parse(path)
 	args := []string{
 		"dependency:tree",
 		"-Djava.net.useSystemProxies=true",
 		fmt.Sprintf("-DoutputFile=%s", f.Name()),
 	}
+	if pom.Modules != nil {
+		args = append([]string{"compile"}, args...)
+	}
+
 	if p.mvnSettingsFile != "" {
 		args = append(args, "-s", p.mvnSettingsFile)
 	}
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -10,11 +10,9 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"strings"
 
-	"github.com/antchfx/xmlquery"
 	"github.com/konveyor/analyzer-lsp/engine/labels"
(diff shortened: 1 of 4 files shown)


shawn-hurley:
Can you help me understand why we want to compile when initializing deps and downloading sources?

jmle:
@shawn-hurley this is for `mvn dependency:sources` to work. When analyzing an application in source mode with multiple submodules which depend on eachother, the command won't work since it won't be able to find some of those dependencies (ie, submodule B which depends on submodule A). You'll get something like this:

```
[ERROR] Failed to execute goal on project daytrader-ee7-web: Could not resolve dependencies for project net.wasdev.wlp.sample:daytrader-ee7-web:war:1.0-SNAPSHOT: Could not find artifact net.wasdev.wlp.sample:daytrader-ee7-ejb:jar:1.0-SNAPSHOT -> [Help 1]```

**Comment by shawn-hurley** [comment]
Is there a way to detect that and then re-attempt w/ that or do we expect the submodule case is the primary case?

File: provider/internal/java/dependency.go [hunk 1]
```diff
@@ -97,52 +96,65 @@ func getMavenLocalRepoPath(mvnSettingsFile string) string {
...
+				version = pom.Properties.Entries[version]
+				if version != "" {
+					dep.Version = version
+				}
+			} else {
+				dep.Version = *d.Version
 			}
-			dep.Name = node.InnerText()
-		} else if node.Data == "artifactId" {
-			dep.Name += "." + node.InnerText()
```
jmle, line 133: Not sure about why this is concatenating with `.` instead of `:`?

provider/internal/java/dependency.go
```diff
@@ -97,52 +96,65 @@ func getMavenLocalRepoPath(mvnSettingsFile string) string {
...
-		} else if node.Data == "version" {
-			dep.Version = node.InnerText()
 		}
-		// Ignore the others
-	}
-	if !reflect.DeepEqual(dep, provider.Dep{}) {
```
jmle, line 139: Also not sure about what this does

shawn-hurley: This checks if the dep that we have determined is empty, if it is n't empty then we add it to the list.

provider/internal/java/dependency.go
```diff
@@ -97,52 +96,65 @@ func getMavenLocalRepoPath(mvnSettingsFile string) string {
...
+		pomDeps = append(pomDeps, *pom.Dependencies...)
+	}
+	if pom.DependencyManagement != nil {
+		pomDeps = append(pomDeps, *pom.DependencyManagement.Dependencies...)
+	}
+
+	// add each dependency found
+	for _, d := range pomDeps {
+		dep := provider.Dep{}
+		dep.Name = fmt.Sprintf("%s.%s", *d.GroupID, *d.ArtifactID)
```
shawn-hurley, line 123: Are there times when GroupID or ArtifactID could be nil, this will panic if they can be.

provider/internal/java/dependency.go
```diff
@@ -97,52 +96,65 @@ func getMavenLocalRepoPath(mvnSettingsFile string) string {
...
-				deps = append(deps, dep)
-				dep = &provider.Dep{}
+
+	// have to get both <dependencies> and <dependencyManagement> dependencies (if present)
+	var pomDeps []gopom.Dependency
+	if pom.Dependencies != nil {
+		pomDeps = append(pomDeps, *pom.Dependencies...)
+	}
+	if pom.DependencyManagement != nil {
+		pomDeps = append(pomDeps, *pom.DependencyManagement.Dependencies...)
```
shawn-hurley, line 117: Can `Dependencies` be nil?

provider/internal/java/dependency.go
```diff
@@ -97,52 +96,65 @@ func getMavenLocalRepoPath(mvnSettingsFile string) string {
...
+	}
+
+	// add each dependency found
+	for _, d := range pomDeps {
+		dep := provider.Dep{}
+		dep.Name = fmt.Sprintf("%s.%s", *d.GroupID, *d.ArtifactID)
+		if *d.Version != "" {
```
shawn-hurley, line 124: @fabianvf wonder if this will make the version parser better?

provider/internal/java/dependency.go
shawn-hurley: goal is to make dep name like, `com.foo.bar.artificatID`

provider/internal/java/dependency.go
jmle: Not with a well formed pom - if it can be compiled, dependencies will have at least groupId and artifactId

provider/internal/java/dependency.go
jmle: Yep, fixed

provider/internal/java/dependency.go
```diff
@@ -97,52 +96,67 @@ func getMavenLocalRepoPath(mvnSettingsFile string) string {
...
-	}
-	if !reflect.DeepEqual(dep, provider.Dep{}) {
-		dep.Labels = []string{fmt.Sprintf("%v=%v", provider.DepSourceLabel, javaDepSourceInternal)}
-		deps = append(deps, dep)
+		deps = append(deps, &dep)
 	}
+
 	m := map[uri.URI][]*provider.Dep{}
-	m[file] = deps
+	m[uri.File(path)] = deps
```
shawn-hurley, line 141: I am a little worried about this change having side effects. is there a reason to move to a URI instead of the file path that is passed in?

provider/internal/java/dependency.go
```diff
@@ -97,52 +96,67 @@ func getMavenLocalRepoPath(mvnSettingsFile string) string {
...
-	path := p.findPom()
-	file := uri.File(path)
+func (p *javaServiceClient) GetDependenciesFallback(ctx context.Context, location string) (map[uri.URI][]*provider.Dep, error) {
+	deps := []*provider.Dep{}
 
-	absPath, err := filepath.Abs(path)
-	if err != nil {
-		return nil, err
+	path := p.findPom()
+	if location != "" {
```
pranavgaikwad, line 103: Can we use `filepath.Abs()` on path?

pranavgaikwad (changes requested):

provider/internal/java/dependency.go
jmle: This is the same, I just removed `file := uri.File(path)` [above](https://github.com/konveyor/analyzer-lsp/pull/368/files#diff-b70ad3b19e5302fa83ea4cbeae98cbd7bb689d2d888102229fe3f133f771f44eL103).

provider/internal/java/dependency.go
jmle: Done :+1:

pranavgaikwad (approved):

PR #370, opened 2023-10-11
:seedling: Add Dockerfile for debugging

jmle:


pranavgaikwad (changes requested):
Can you please add a doc that explains usage of this? and perhaps, move this into hack/ directory or something?

jmle:
@pranavgaikwad do you think it would make sense to have a `docker` directory with all the different docker images?

pranavgaikwad:
@jmle We have the release automation working off of the ./Dockerfile. It's good to have the default ones in the root path I think.

jmle:
@pranavgaikwad moved to `debug` and provided a README

jmle:
@pranavgaikwad I added the requested changes - I think this can be useful, especially for people starting with the project

PR #372, opened 2023-10-11
:bug: Fix ppc64le build

JonahSussman:
Added gcc-c++ to the Dockerfile so python-lsp can build. Most likely a stopgap solution until we get better management of lsp servers.


JonahSussman:
[Build fixed on my branch](https://github.com/JonahSussman/analyzer-lsp/actions/runs/6488847607/job/17622297540), with the only failing part being "push to quay"

pranavgaikwad (approved):

PR #382, opened 2023-10-18
:bug: Fix "grpc unable to get info" messages at startup 

### LINKED ISSUE #343: Fix "grpc unable to get info" messages at startup
When running the tool, we always get failed to connect errors. Something along the lines of this:

```
time="2023-09-27T04:36:15Z" level=error msg="grpc unable to get info" error="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp [::1]:33925: connect: connection refused\"" logger=go provider=grpc
```

JonahSussman:
Fixes #343

Note, that there's some issue where the java provider is matching comments. @shawn-hurley and @pranavgaikwad  I think this related to something you guys were talking about earlier.

### DIFF
--- a/provider/grpc/provider.go
+++ b/provider/grpc/provider.go
@@ -55,7 +55,9 @@ func (g *grpcProvider) ProviderInit(ctx context.Context) error {
 }
 
 func (g *grpcProvider) Capabilities() []provider.Capability {
-	r, err := g.Client.Capabilities(context.TODO(), &emptypb.Empty{})
+	// Capabilities requests happen at the start of the analyzer-lsp, so it makes
+	// sens to wait until the connection is ready.
+	r, err := g.Client.Capabilities(context.TODO(), &emptypb.Empty{}, grpc.WaitForReady(true))
 	if err != nil {
 		// Handle this smarter in the future, for now log and return empty
 		g.log.V(5).Error(err, "grpc unable to get info")


pranavgaikwad (approved):

provider/grpc/provider.go
```diff
@@ -55,7 +55,9 @@ func (g *grpcProvider) ProviderInit(ctx context.Context) error {
 }
 
 func (g *grpcProvider) Capabilities() []provider.Capability {
-	r, err := g.Client.Capabilities(context.TODO(), &emptypb.Empty{})
+	// Capabilities requests happen at the start of the analyzer-lsp, so it makes
+	// sens to wait until the connection is ready.
+	r, err := g.Client.Capabilities(context.TODO(), &emptypb.Empty{}, grpc.WaitForReady(true))
```
shawn-hurley, line 60: I think that if you use wait for ready, that you need to set a timeout 

```. Otherwise, the RPC client will block the call until a connection is available (or the call is canceled or times out) and will retry the call if it fails due to a transient error.```

I also wonder if making the log a deeper verbosity and keeping the call as is, would allow us to debug the case when the server is not ready easier.

Just some thoughts and questions WDYT?

shawn-hurley (changes requested):
I think that we either need a timeout or to change the logging verbosity but we need to make one of the two changes IMO.

provider/grpc/provider.go
JonahSussman: Didn't see the blocking thing with `WaitForReady`, I'll add a timeout! Technically, this could be solved by putting a `time.sleep` just before calling this, but I wanted a more robust solution. I also agree with making the verbosity deeper

I couldn't find a way to wait/check for the grpc server to be ready after instantiation. I think that'd probably be the best way to do it if we could find it

shawn-hurley: The timeout can be added to the context.

I would probably prefer to use our retry logic (that calls the caps and the reason you get this message every time) and reduce the logging verbosity. This will allow us to turn up the verbosity and see if the server never becomes active. WDYT?

JonahSussman:
Closing PR for now, the "grpc unable to get info" message isn't actually that intrusive for now

PR #385, opened 2023-10-23
:bug: Fix dependency handling

### LINKED ISSUE #384: [BUG] Dependencies not properly treated in multi-module applications

### Konveyor version

0.3.0-beta.1.1

### Priority

Major

### Current Behavior

In multi module applications, dependencies are not being properly fetched.

### Expected Behavior

The dependencies view should show all direct and transitive dependencies.

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Analyze with konveyor the[ "petclinic" application](https://github.com/spring-projects/spring-petclinic).
2. Go to dependencies.
3. See that there are many dependencies missing.

### Environment

```markdown
- Fedora 38
- Konveyor beta1.1
```

### Anything else?

_No response_

jmle:
Fixes #384 

- Handle multimodule project dependencies properly
- Only decompile dependencies whose source hasn't been found

Before:
image

After:
image

Fixes https://github.com/konveyor/analyzer-lsp/issues/384

### DIFF
--- a/demo-dep-output.yaml
+++ b/demo-dep-output.yaml
@@ -894,18 +894,202 @@
 - fileURI: file:///analyzer-lsp/examples/java/pom.xml
   provider: java
   dependencies:
-  - name: io.fabric8.kubernetes-client-api
-    version: 6.0.0
+  - name: com.fasterxml.jackson.core.jackson-core
+    version: 2.13.3
     type: compile
-    resolvedIdentifier: 3f54cdb10f54b413fe4b8a0d4d044d33174bd271
+    resolvedIdentifier: a27014716e4421684416e5fa83d896ddb87002da
     labels:
     - konveyor.io/dep-source=open-source
     - konveyor.io/language=java
-    prefix: file:///root/.m2/repository/io/fabric8/kubernetes-client-api/6.0.0
+    prefix: file:///root/.m2/repository/com/fasterxml/jackson/core/jackson-core/2.13.3
+  - name: com.fasterxml.jackson.core.jackson-databind
+    version: 2.13.3
+    type: compile
+    resolvedIdentifier: 56deb9ea2c93a7a556b3afbedd616d342963464e
+    labels:
+    - konveyor.io/dep-source=open-source
+    - konveyor.io/language=java
+    prefix: file:///root/.m2/repository/com/fasterxml/jackson/core/jackson-databind/2.13.3
+  - name: com.fasterxml.jackson.datatype.jackson-datatype-jsr310
+    version: 2.13.3
+    type: compile
+    resolvedIdentifier: ad2f4c61aeb9e2a8bb5e4a3ed782cfddec52d972
+    labels:
+    - konveyor.io/dep-source=open-source
+    - konveyor.io/language=java
+    prefix: file:///root/.m2/repository/com/fasterxml/jackson/datatype/jackson-datatype-jsr310/2.13.3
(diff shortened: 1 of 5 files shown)


demo-dep-output.yaml
```diff
@@ -894,257 +894,233 @@
 - fileURI: file:///analyzer-lsp/examples/java/pom.xml
   provider: java
   dependencies:
-  - name: io.fabric8.kubernetes-client-api
-    version: 6.0.0
+  - name: com.fasterxml.jackson.core.jackson-core
```
shawn-hurley, line 897: There seems to be a new order happening. I don't think that I see new entries here. Can you help me understand if this is intended, and does it break any other tests?

provider/internal/java/dependency.go
```diff
@@ -195,7 +195,6 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
 	args := []string{
 		"dependency:tree",
 		"-Djava.net.useSystemProxies=true",
-		fmt.Sprintf("-DoutputFile=%s", f.Name()),
```
shawn-hurley, line 198: If you are not putting it out to a file, then I think we should be able to remove the creation of the temp file and the defer from above.

provider/internal/java/dependency.go
```diff
@@ -204,30 +203,34 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
...
 	// get the graph output
 	cmd := exec.Command("mvn", args...)
 	cmd.Dir = moddir
-	err = cmd.Run()
+	mvnOutput, err := cmd.CombinedOutput()
 	if err != nil {
 		return nil, err
 	}
 
+	if _, err = f.Write(mvnOutput); err != nil {
```
shawn-hurley, line 215: I don't know why we would write and then read from the file for it to be removed. I feel like remove everything to do with the file.

provider/internal/java/dependency.go
```diff
@@ -241,6 +244,37 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
...
+	endRegex := regexp.MustCompile(`\[INFO\] -*$`)
+
+	submod := 0
+	gather, skipmod := false, true
+	for _, line := range lines {
+		if beginRegex.Find([]byte(line)) != nil {
+			gather = true
+			submoduleTrees = append(submoduleTrees, []string{})
+		} else if gather {
+			if endRegex.Find([]byte(line)) != nil {
```
shawn-hurley, line 261: nit: use continue instead of else blocks 

It will make the for loop more clear.

provider/internal/java/provider.go
```diff
@@ -346,19 +348,33 @@ func resolveSourcesJars(ctx context.Context, log logr.Logger, location, mavenSet
 	}
 	cmd := exec.CommandContext(ctx, "mvn", args...)
 	cmd.Dir = location
-	err = cmd.Run()
+	output, err := cmd.CombinedOutput()
 	if err != nil {
 		return err
 	}
+
+	if _, err = mvnOutput.Write(output); err != nil {
```
shawn-hurley, line 356: Same comments on the file IMO

shawn-hurley (commented):
Minor concerns, and I just want to understand the ordering after this change.

pranavgaikwad (approved):

Issue #393, opened 2023-10-31
[BUG] Java / Maven analysis does not work if the project does not compile with JDK 17


ascheman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

60c4fcbe62658708fb4290cbdcb24b75e9619814

### Priority

Blocker

### Current Behavior

If a Maven project does not compile with JDK 17, the analysis will abort.

### Expected Behavior

Make it work with misc. Java / JDK versions. Perhaps some hint for the desired Java version is necessary?

### How Reproducible

Always (Default)

### Steps To Reproduce

Project which is incompatible with JDK 17 needed, e.g., by calling `sun.misc.Unsafe`

### Anything else?

If the call to Maven fails, it would be great to get the (full) output of the background process for failure analysis.

ascheman:
Check [Slack](https://kubernetes.slack.com/archives/C04QZJFQ0UA/p1698772458802029?thread_ts=1698750038.252959&cid=C04QZJFQ0UA) for some discussion

ascheman:
The compile goal can be dropped from the Maven call when using a newer version of Maven (locally tested with Maven 3.9.5). According to @jmle the compile goal needed to be added as Maven was not properly analysing dependencies for modules in a multi-module build. Of course, this didn't work with Maven 3.6.3 (as installed by [Java-Analyzer-Bundle](https://github.com/konveyor/java-analyzer-bundle/blob/9cb5a19d87531487df378afbb37de4c60d16ae87/Dockerfile#L11)).

shawn-hurley:
I worry about trying to use a version that is un-installable in UBI9. Not to say we can't figure it out, but did want to note the limitation that we would have

This is an excellent thing for us to look in. Is there a specific version or syntax you are hitting for my edification,

ascheman:
I don't understand the limitation? You cannot install it with `dnf`? 
Then why do you [setup python packages in other places](https://github.com/ascheman/analyzer-lsp/blob/main/Dockerfile#L25-L25)? Isn't this also beyond UBI installation processes?

Regarding Maven/Dependency computation: Find the maximum version which is possible (perhaps it is not required to get the latest 3.9.x, maybe 3.8.x or something will do as well)? Perhaps it is even possible to use Maven 3.6 and explicitely use the dependency plugin in version 3.x?

[...]

Issue #392, opened 2023-10-31
[BUG] Maven Dependencies are not (always) correctly parsed


ascheman:
When analyzing Maven output, the project has to cope with different formatting.

# Coloured output

When run interactively (even when started from a local analyser execution in the background),
Maven may use  color highlighting (Terminal specific escape sequences) which will result in parsing problems.

# Maven Version Problem

Depending on the Maven version, the output of `dependency:tree` is different, e.g.,

* Maven 3.6.3 outputs each lifecycle execution with `maven-dependency-plugin:<version>:tree`, while
* Maven 3.9.5 outputs `dependency:<version>:tree` at the same position.

# Maven dependency string splitting problem

Depending on the type of dependency, the dependency plugin prints differently composed strings, e.g.,

* `<groupId>:<artifactId>:<packaging>:<version>` vs.
* `<groupId>:<artifactId>:<packaging>:<classifier>:<version>`.

jmle:
@pranavgaikwad @ascheman I think these issues are probably solved now, right?

ascheman:
Hmm, I would appreciate if you could either cherry-pick my changes or at least refer to the respective issue (this one) in your commit messages. 
I was not aware my reported bugs are (partially) already resolved by your changes, e.g. d2956153fc84e43fb05a8bec96f53061649e698a.
I didn't find much time to work on them myself in the last days.
But I have them on my agenda and spent some time on them today, only to find them partially fixed already in your code by now (without any hint to the respective issue).

Currently, I cannot see how you handled different dependency results (split by ':') correctly, as I suggest it here: https://github.com/qaware/analyzer-lsp/commit/541a2d097353fe72827ce573d7d15353d3920e84?
I wanted to provide some tests for these cases, but I am unsure whether you are working in parallel on this? We should avoid such situations (as I guess we both have enough to do).

Additionally I tried to provide a smaller fix of some logging under the shield of this issue, i.e., https://github.com/qaware/analyzer-lsp/commit/9465c3aa5d66ba43c58b8993a2f2a38081e1ef6e.
Will you pick this up also?

jmle:
[...]

Issue #391, opened 2023-10-31
RfE: Dependency Analyzer should enable log levels


ascheman:
As a user of the dependency analyzer tool I would like to be 
able to specify a `--verbose <level>` like it is possible with
the full-blown `konveyor-analyzer`.

Issue #390, opened 2023-10-31
[BUG] Property resolution leads to SEGV when POM has no own props


ascheman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

How can I get it? I am using Analyzer-LSP directly!

### Priority

Blocker

### Current Behavior

When analysing a Java POM with a dependency as a property, analyzer tries to resolve the property in the same POM. If the POM contains no properties at all, this leads to a segmentation violation (in [line 141 of `provider/internal/java/dependeny.go`](https://github.com/konveyor/analyzer-lsp/blob/9369083d47c8b056775300b38adeae856612f397/provider/internal/java/dependency.go#L141)):

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5b3ddc]
```

### Expected Behavior

Try to resolve property via parent POM (hierarchy) or leave unresolved otherwise (which [is the default, if the property cannot be found in an existing POM property list](https://github.com/konveyor/analyzer-lsp/blob/9369083d47c8b056775300b38adeae856612f397/provider/internal/java/dependency.go#L142)).

### How Reproducible

Always (Default)

### Steps To Reproduce

Prepare a project with POM like this.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <groupId>some-project</groupId>
    <artifactId>some-artifact</artifactId>
    <version>does-not-matter</version>
    
    <dependencies>
        <dependency>
            <groupId>some-dep-group</groupId>
            <artifactId>some-dep-artifact</artifactId>
            <version>${cannot-be-resolved}</version>
        </dependency>
    </dependencies>
</project>
``` 

And analyze it.

### Environment

```markdown
Have to provide a minimal example ...
```

jmle:
[...]

PR #394, opened 2023-10-31
:bug: Misc. first fixes for 390, 391, 392, 393

ascheman:
Opening this as a draft for all of my work regarding #390, #391, #392, and #393 to enable further discussions.

### DIFF
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -201,17 +201,9 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
 
 	moddir := filepath.Dir(path)
 
-	pom, err := gopom.Parse(path)
-	if err != nil {
-		return nil, err
-	}
-
 	args := []string{
 		"dependency:tree",
 		"-Djava.net.useSystemProxies=true",
-	}
-	if pom.Modules != nil {
-		args = append([]string{"compile"}, args...)
 	}
 
 	if p.mvnSettingsFile != "" {
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -350,12 +350,32 @@ func (p *javaServiceClient) parseDepString(dep, localRepoPath string) (provider.
 	// Split string on ":" must have 5 parts.
 	// For now we ignore Type as it appears most everything is a jar
 	parts := strings.Split(dep, ":")
-	if len(parts) != 5 {
-		return d, fmt.Errorf("unable to split dependency string %s", dep)
+	if len(parts) >= 3 {
+		// Its always <groupId>:<artifactId>:<Packaging>: ... then
+		d.Type = parts[2]
+		if len(parts) == 6 {
+			// parts[5] is the <scope>
+			// parts[3] is the <classifier>
+			d.Version = parts[4]
+		} else if len(parts) == 5 {
+			// parts[4] is the <scope>
+			d.Version = parts[3]
+		} else {
+			p.log.Info("Cannot derive version from dependency string", "dependency", dep)
+			d.Version = "Unknown"
+		}
+	} else {
+		p.log.Info("unable to analyze dependency", "dependency", dep)
+		d.Name = "Unknown"
+		d.Version = "Unknown"
+		d.Type = "Unknown"
+		d.Labels = []string{""}
+		d.FileURIPrefix = ""
+		d.ResolvedIdentifier = ""
+
+		return d, nil
 	}
 	d.Name = fmt.Sprintf("%s.%s", parts[0], parts[1])
-	d.Version = parts[3]
-	d.Type = parts[4]
 
 	fp := filepath.Join(localRepoPath, strings.Replace(parts[0], ".", "/", -1), parts[1], d.Version, fmt.Sprintf("%v-%v.jar.sha1", parts[1], d.Version))
 	b, err := os.ReadFile(fp)
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -202,6 +202,7 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
 	moddir := filepath.Dir(path)
 
 	args := []string{
+		"-B",
 		"dependency:tree",
 		"-Djava.net.useSystemProxies=true",
 	}
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -109,6 +109,9 @@ func (p *javaServiceClient) GetDependenciesFallback(ctx context.Context, locatio
 		path = location
 	}
 	pom, err := gopom.Parse(path)
+	p.log.V(10).Info("Analyzing POM",
+		"POM", fmt.Sprintf("%s:%s:%s", pomCoordinate(pom.GroupID), pomCoordinate(pom.ArtifactID), pomCoordinate(pom.Version)),
+		"error", err)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


provider/internal/java/util.go
```diff
@@ -127,10 +127,10 @@ func decompile(ctx context.Context, log logr.Logger, filter decompileFilter, wor
 					continue
 				}
 				cmd := exec.CommandContext(
-					ctx, "java", "-jar", "/bin/fernflower.jar", job.inputPath, outputPathDir)
+					ctx, "java", "-jar", "/usr/local/bin/fernflower.jar", job.inputPath, outputPathDir)
```
pranavgaikwad, line 130: This will require us to fix [the Dockerfile in https://github.com/konveyor/java-analyzer-bundle so that ](https://github.com/konveyor/java-analyzer-bundle/blob/9cb5a19d87531487df378afbb37de4c60d16ae87/Dockerfile#L35), is his change an abosolute requirement?

ascheman: Correct, but it would make decompilation more portable as pointed out in b52929fb86eb7d455e9acf4af97b13cd7de40f09. I could not test with this path on MacOS, as /bin is write protected (for good reasons).

In general I would appreciate if such paths were configurable by file or environment. We could do this in the long run. 

Having said that, my inner monk ruffles up at the thought of having a jar file in the /bin folder (even if it's only a container, and even if the file is executable), as the /bin folder traditionally is the core folder for OS binaries. User add-ons should go to /usr/local or /opt.

jmle: I see that the main reason for this change is debugging. I'll send you some indications about how to debug from outside the container. I don't object to this change, but bare in mind that the analyzer will mostly be used from within the konveyor ecosystem in a containerized way, and that is the path we're mainly following.

ascheman: I fully understand that the end-user will mostly work with containers (which is perfectly fine for me) in order to avoid tons of local configurations, executables/tools etc. But to be honest, as a newbie development contributor, it is hard to follow all of these dependencies, and even harder to maintain them. 

Maybe for this change it is not much of a problem. But check out the change of Maven to 3.9 to avoid the compile goal (#393): 
* To get it running, 
  * you need to change the base image with the Java/Maven installations. 
  * Then build the analyzer-lsp image based on that. 
  * Finally build the Kantra image. 
* During runtime, 
  * `kantra` calls `konveyor-analyzer` (and some other tools), 
  * which call `mvn` at some point. 
If something goes wrong, it is hard to debug as the tools simply abort without leaving much trace to the root cause (perhaps I need to checkout the Jaeger integration?). I never saw that Maven failed (or even why) until I started the debugger.

To me it is much easier to run/debug all of this locally and have full control over each of the components.
Hence, I have the strong wish to be able to work locally with some (well-documented!) prerequisites as well as in containers. Setup should support both modes in my opinion.

BTW: Even when working completely in containers (as the end-user will do) it is essential to have a thorough documentation of all components, their workflows, all (intermediate) artifacts, and in particular how to observe their behaviour (logging/tracing/monitoring). Sorry, for repeating this over and over again ;-)

provider/internal/java/dependency.go
```diff
@@ -138,9 +141,20 @@ func (p *javaServiceClient) GetDependenciesFallback(ctx context.Context, locatio
 		if *d.Version != "" {
 			if strings.Contains(*d.Version, "$") {
 				version := strings.TrimSuffix(strings.TrimPrefix(*d.Version, "${"), "}")
-				version = pom.Properties.Entries[version]
-				if version != "" {
+				p.log.V(10).Info("Searching for property in properties",
+					"property", version,
+					"properties", pom.Properties)
+				if pom.Properties == nil {
+					p.log.Info("Cannot resolve version property value as POM does not have properties",
```
shawn-hurley, line 148: very very small nit: what do we think of making the this `..."groupID", pomCo..., "artifactID", pomCor....)` 

Just wondering what folks think

provider/internal/java/dependency.go
```diff
@@ -336,22 +350,47 @@ func (p *javaServiceClient) parseDepString(dep, localRepoPath string) (provider.
...
-	d.Version = parts[3]
-	d.Type = parts[4]
 
 	fp := filepath.Join(localRepoPath, strings.Replace(parts[0], ".", "/", -1), parts[1], d.Version, fmt.Sprintf("%v-%v.jar.sha1", parts[1], d.Version))
-	b, err := os.ReadFile(fp)
-	if err != nil {
-		// Log the error and continue with the next dependency.
-		p.log.V(5).Error(err, "error reading SHA hash file for dependency", "dep", d.Name)
-		// Set some default or empty resolved identifier for the dependency.
+	if _, err := os.Stat(fp); os.IsNotExist(err) {
```
shawn-hurley, line 381: Will ReadFile, give you the same error, or does it give a different error?

provider/internal/java/util.go
shawn-hurley: While I agree in principle, the word support here is something to consider. 

Because the setup for the components is pretty complex, I worry about supporting everyone's local environments. This could be a long tail that would cost a lot of cycles, IMO. 

I would like to know if we can get an agreement on documentation and best effort to answer questions when testing locally, but there are no guarantees on time/fixes vs. in the container where there should be guarantees on this. Thoughts?

shawn-hurley: Oh, and in addition, I am plus on on moving things out of `/bin/` that seems like a good citizen thing to do.

provider/internal/java/dependency.go
ascheman: ?? ReadFile returns an error if the file does not exist. Therefore, I consider it a good practice in programming to check whether the file exists, before trying to open it.

shawn-hurley: If that is the case, I would leave this logic as is. 

We get the same error, and it goes to the log explaining that the file does not exist. And we remove complexity to only have a single if/else block.

ascheman: Seems we have different opinions here. But may I remind you of the _good-citizen_ practice mentioned elsewhere? ;-)

provider/internal/java/util.go
ascheman: Starting a smooth migration to better local executability / improving abilities to debug does not necessarily mean, that all possible paths must be fully supported. In this case, the change is invisible to almost everyone but makes life easier for someone developing outside of containers while not harming anyone else.

provider/internal/java/dependency.go
ascheman: The pomCoordinate function was just a quick hack to cope with strings which are either set or "". Any better approach is highly appreciated.

provider/internal/java/dependency.go
shawn-hurley: Please see the library function ReadFile:
 https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/os/file.go;drc=79a8997ac5592f11388d9f26ab3fff65ec876dbd;l=714

provider/internal/java/dependency.go
shawn-hurley: Oh, I was not suggesting changing the helper function, I think that for this particular thing, it is a nice UX boost. I was discussing of making the log statement use the key, value pairs for groupID and artifactID instead of combining in a single key(POM). Just think it might be cleaner.

provider/internal/java/dependency.go
fabianvf: I think in golang the typical practice would be to attempt to read the file and check the error for failure. 

As far as I can tell there are no differences between the approaches other than added handling complexity and an extra syscall if we check for existence first. The ReadFile function is built with the expectation that it may be called on non-existent files and has no unpleasant side-effects. 

Since we log the error anyway it will be clear the reason the file read failed without the extra check

pranavgaikwad: I think what shawn's pointing at is that - it's redundant to check for file exists, as the `err` variable in ReadFile will capture the same error itself.

pranavgaikwad: oops, sorry, fabian and I responded at the same time, +1

ascheman: Will not push this anymore. Nevertheless, an error is logged for missing .sha1 files which might not be in the local repo for my own modules. What bothers me here, is that this is logged as error. Errors imply that I (as a user/developer/ops person) have done something wrong. And there is no fault in this case. So for my own modules I would like to see the message logged in maximum as a warning, or even only info or lower (debug).

shawn-hurley: I think I agree with this and would plus one logging this as a V(5).Info while including the error as a key-value pair. You bring up a very good point about it being an error.

provider/internal/java/dependency.go
```diff
@@ -336,22 +350,47 @@ func (p *javaServiceClient) parseDepString(dep, localRepoPath string) (provider.
...
+		p.log.V(5).Info("File does not exist", "file", fp)
 		d.ResolvedIdentifier = ""
 	} else {
-		d.ResolvedIdentifier = string(b)
+		b, err := os.ReadFile(fp)
+		if err != nil {
+			// Log the error and continue with the next dependency.
+			p.log.V(5).Error(err, "error reading SHA hash file for dependency", "dep", d.Name)
```
shawn-hurley, line 388: To be clear add an if statement here: `if os.IsNotExists(err) { p.log.V(5).Info("file does not exists", "file", fp, "err", err) } else { p.log.V(5).Error(err, "error reading SHA hash file for dependency", "dep", d.Name) } `

PR #397, opened 2023-11-02
:bug: Fix compile step

### LINKED ISSUE #393: [BUG] Java / Maven analysis does not work if the project does not compile with JDK 17

jmle:
Fixes #393 

Remove "compile" goal from maven invocations, as some applications may not be compilable with the JDK version we bundle.

Related to https://github.com/konveyor/analyzer-lsp/issues/393 and https://github.com/konveyor/analyzer-lsp/issues/393

### DIFF
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -180,17 +180,9 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
 
 	moddir := filepath.Dir(path)
 
-	pom, err := gopom.Parse(path)
-	if err != nil {
-		return nil, err
-	}
-
 	args := []string{
 		"dependency:tree",
 		"-Djava.net.useSystemProxies=true",
-	}
-	if pom.Modules != nil {
-		args = append([]string{"compile"}, args...)
 	}
 
 	if p.mvnSettingsFile != "" {
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -224,7 +224,7 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
 func extractSubmoduleTrees(lines []string) [][]string {
 	submoduleTrees := [][]string{}
 
-	beginRegex := regexp.MustCompile(`maven-dependency-plugin:[\d\.]+:tree`)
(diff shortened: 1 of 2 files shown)


provider/internal/java/dependency.go
```diff
@@ -180,18 +180,10 @@ func (p *javaServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI
 
 	moddir := filepath.Dir(path)
 
-	pom, err := gopom.Parse(path)
```
shawn-hurley, line 183: Does this remove that dep too?

shawn-hurley:
Do we have a multi-module test case yet? I would love to see that added with/ this change to make sure we don't break that in the future (I am worried about me breaking it :) )

provider/internal/java/dependency.go
pranavgaikwad: think its still used in fallback

pranavgaikwad (commented):
Does this completely fix #392?

pranavgaikwad:
> Do we have a multi-module test case yet? I would love to see that added with/ this change to make sure we don't break that in the future (I am worried about me breaking it :) )

Captured in issue here, will add sometime this week https://github.com/konveyor/analyzer-lsp/issues/399

ascheman:
> Do we have a multi-module test case yet? I would love to see that added with/ this change to make sure we don't break that in the future (I am worried about me breaking it :) )

If it helps: The problem came up in my (company project) multi project build. Tested the patch with that project already.

pranavgaikwad (approved):

provider/internal/java/dependency.go
jmle: Yep, still used somewhere else

ascheman:
Thanks!

Issue #406, opened 2023-11-06
[BUG] Decompilation of unresolved dependency sources is not working as expected


pranavgaikwad:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.3.0-beta.2

### Priority

Critical

### Current Behavior

In the tackle-testapp test, when you select `cloud-readiness` target, the analyzer is supposed to decompile `config-utils` library which is not being deompiled as expected because there's a bug in reading the `dependency:sources` output. This causes the `local-storage-00001` rule to not match. 

### Expected Behavior

The issue for rule `local-storage-0001` must be matched returning results for matches from `config-utils` library. 

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Run analysis for `tackle-testapp` with the correct maven settings. 
2. Select targets `cloud-readiness` 
3. Check for issue generated for `local-storage-00001` rule with title `File System - Java IO`

### Environment

```markdown
- **Analyzer Image SHA** 
- **JDTLS  base image SHA**
- **Provider Settings JSON**
```

pranavgaikwad:
This is still not fixed in some scenarios :(

PR #409, opened 2023-11-07
:bug: Fix source decompilation

jmle:
- Only decompile NOT resolved which don't appear in the RESOLVED list.
- Do not rely on sources classifier
- Maven to output raw (without colors)
- Uses fixed version of the maven-dependency-plugin. This is a temporary fix, until we can get the fix into the plugin and use official versions again.

Depends on https://github.com/konveyor/java-analyzer-bundle/pull/81
Fixes https://github.com/konveyor/analyzer-lsp/issues/406

### DIFF
--- a/provider/internal/java/dependency.go
+++ b/provider/internal/java/dependency.go
@@ -247,7 +247,14 @@ func extractSubmoduleTrees(lines []string) [][]string {
 				continue
 			}
 
-			line = strings.TrimLeft(line, "[INFO] ")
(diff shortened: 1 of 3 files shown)


pranavgaikwad:
Hey folks, I tested this PR with tackle-testapp and daytrader apps, this fixes the issues once and for all. Since I am the co-author, I'd like someone else to take a look as well.

pranavgaikwad (approved):

fabianvf (approved):
lgtm

PR #411, opened 2023-11-07
Build the addon image so analyzer code is actually tested by global CI

fabianvf:


shawn-hurley:
Should this happen in the add-on on merge?

fabianvf:
@shawn-hurley what do you mean?

shawn-hurley:
I see, because the add-on uses our built image as the base? 

Sorry that makes total sense if that is the case I am sorry

fabianvf:
@shawn-hurley yeah exactly, this PR just actually puts the analyzer into the flow by ensuring that the addon image is rebuilt on top of the changes from the PR. A little fragile/interconnected for my tastes but probably worth it to get the tests running.

fabianvf:
The failure we're getting right now seems legitimate, if unrelated to the analyzer

pranavgaikwad (approved):

PR #412, opened 2023-11-09
:sparkles: add dependency location resolver and get code locations for XML, JSON files 

### LINKED ISSUE #254: Add code snippets to JSON and XML in builtin provider
It'd be great to have code locations in JSON and XML capabilities too similar to what I did in #253

### LINKED ISSUE #410: Add code snippet capability to things coming from a dependency condition
At the time of creating dependencies, we can use the Extras field on a Dep to communicate line numbers, we already have a file URI. This could be as easy as grepping the dependency name intelligently to get the line number, given that deps would be unique in a pom and chances of getting a wrong line number from grep are pretty low.

pranavgaikwad:
Fixes #410 
Fixes #254 

There were two choices to do this: 

1. Add a new optional field lineNumber to Dep type, and let the providers populate it at the time of dependency discovery. This is straight-forward and probably preferred, but requires a change in the output API.  
2. Add a new interface like Snipper, that providers can implement for deps to get locations for a certain dependency. 

I chose No. 2 because output API change not needed. But I am not fan of it as its complicated & perf penalty. Asking for your inputs on the approach. If we do go with 2, I think we should we should still remove it in later versions in favor of a new optional field on Dep. Even if we did add it now, it would be non-intrusive.


pranavgaikwad:
@shawn-hurley @fabianvf @djzager Let me know what you guys think of this approach

provider/internal/java/dependency.go
```diff
@@ -517,3 +539,75 @@ func loadDepLabelItems(r io.Reader, depToLabels map[string]*depLabelItem, label
...
+			StartPosition: engine.Position{
+				Line: val,
+			},
+			EndPosition: engine.Position{
+				Line: val,
+			},
+		}, nil
+	}
+	defer func() {
+		j.depsLocationCache[cacheKey] = location.StartPosition.Line
```
shawn-hurley, line 578: do we have any issues here with  threading and need a mutex?

provider/internal/java/dependency.go
```diff
@@ -517,3 +539,75 @@ func loadDepLabelItems(r io.Reader, depToLabels map[string]*depLabelItem, label
...
+	}()
+
+	groupId := dep.Extras[groupIdKey]
+	artifactId := dep.Extras[artifactIdKey]
+	path := dep.Extras[pomPathKey]
+	cmd := exec.CommandContext(ctx, "pcre2grep", []string{
```
shawn-hurley, line 584: Why can't we use grep -E?

shawn-hurley (approved):
This looks good to me, did have a couple of concerns but overall I like the cache implementation

provider/internal/java/dependency.go
pranavgaikwad: @shawn-hurley main problem with grep is that it doesn't do a multiline search. Or at least, I couldn't find any litarature around that.

provider/internal/java/dependency.go
pranavgaikwad: probably yes, good catch

provider/provider.go
```diff
@@ -561,17 +566,34 @@ func (dc DependencyCondition) Evaluate(ctx context.Context, log logr.Logger, con
 		return resp, nil
 	}
 
+	var depLocationResolver DependencyLocationResolver
+	if val, ok := dc.Client.(DependencyLocationResolver); ok {
```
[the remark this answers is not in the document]
pranavgaikwad: ahh cool, good catch.

provider/internal/java/dependency.go
shawn-hurley: heard, makes sense, just wondering

pranavgaikwad: @shawn-hurley had to remove usage of `pcregrep` as it is unavailable in ubi. wrote a small parser myself that respects the timeout context

provider/internal/builtin/provider.go
```diff
@@ -128,6 +128,8 @@ func (p *builtinProvider) Init(ctx context.Context, log logr.Logger, config prov
 		config:                             config,
 		tags:                               p.tags,
 		UnimplementedDependenciesComponent: provider.UnimplementedDependenciesComponent{},
+		locationCache:                      make(map[string]float64),
```
shawn-hurley, line 131: Why float and not int?

shawn-hurley (approved):
Looks good except for the one question around float64 and int

provider/internal/builtin/provider.go
pranavgaikwad: For some reason, in provider API, the value is float

fabianvf (approved):

fabianvf:
Looks good!

PR #414, opened 2023-11-09
:bug: Also download deps prior to downloading sources

### LINKED ISSUE #406: [BUG] Decompilation of unresolved dependency sources is not working as expected

pranavgaikwad:
Fixes #406

### DIFF
--- a/provider/internal/java/provider.go
+++ b/provider/internal/java/provider.go
@@ -338,6 +338,20 @@ func resolveSourcesJars(ctx context.Context, log logr.Logger, location, mavenSet
 	}
 
 	args := []string{
+		"dependency:go-offline",
+		"-Djava.net.useSystemProxies=true",
+	}
+	if mavenSettings != "" {
+		args = append(args, "-s", mavenSettings)
+	}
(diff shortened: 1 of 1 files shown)


jmle (approved):
Looks good :+1:

PR #419, opened 2023-11-13
:bug: copy the custom plugin to m2

pranavgaikwad:
API Tests PR: 55

Sorry, `install-file` ended up failing because it wasn't resolving the plugin deps correctly and was creating artifact without full classpath.

### DIFF
--- a/.github/workflows/demo-testing.yml
+++ b/.github/workflows/demo-testing.yml
@@ -56,4 +56,3 @@ jobs:
     uses: konveyor/ci/.github/workflows/global-ci.yml@main
     with:
       component_name: tackle2-addon-analyzer
-    secrets: inherit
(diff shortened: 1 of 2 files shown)


fabianvf (commented):
Not that familiar with what this is actually doing, but the code itself looks sane

shawn-hurley:
Why this over installing with `mvn install-file -s settings.xml` using the settings?

pranavgaikwad:
@shawn-hurley install goal didnt work. I was seeing errors with classpath when running. I saw that there are additional metadata files with the plugin that weren't present when I do the install.

shawn-hurley (approved):

pranavgaikwad:
This is no longer needed as we are using the other plugin now

Issue #423, opened 2023-11-14
[BUG] Dependency plugin's go-offline fails for absent artifacts


jmle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.3.0-beta2

### Priority

Critical

### Current Behavior

Right now, with the custom fixed maven-dependency-plugin, there is still an ongoing issue when trying to download dependencies that cannot be found (absent): the maven run fails if there is a single dependency that is nowhere to be found in maven central or any of the other potentially configured repositories.

### Expected Behavior

Dependencies that cannot be found should be ignored in this case, and the run should continue.

### How Reproducible

Always (Default)

### Steps To Reproduce

This happens as of commit `e73e5df290d7c74aa1e58b60c1c7356252822984`.

### Anything else?

In order to fix this, we need to switch to https://github.com/qaware/go-offline-maven-plugin, which seems to fix all issues. It also provides a simpler way of identifying dependencies that couldn't be found to decompile them. Also, dependencies and their sources can be downloaded using a single command.

PR #426, opened 2023-11-15
:bug: Use qaware's dependency plugin to avoid all issues related to maven dependency plugin

jmle:
API Tests PR: 55

With the new plugin, dependency extraction is simplified to parsing a single line. We also don't need to download dependencies and sources separately; everything is done in a single command. If a dependency is not found in any of the configured repositories, then we won't be able to decompile it (it is absent), therefore we are not interested in those and ignore them.

- Fixes https://github.com/konveyor/analyzer-lsp/issues/423
- We also wouldn't need https://github.com/konveyor/analyzer-lsp/pull/419 anymore


pranavgaikwad (approved):
The e2e test passing is a good sign that this fixes the issues.

PR #427, opened 2023-11-15
:bug: Test cases for #390

ascheman:
Adding test cases for #390.

The Java example was refactored to make it run (i.e., `mvn compile` at least), and extended by a submodule.
The submodule contains a property based dependency version which cannot be immediately resolved, as there is no `properties` definition in the submodule itself. 

The hotfix provided by #425 prevents the binary from failing. 
Nevertheless, the version property should be resolved from the parent pom which is not yet addressed.

As the bug report only was about the SEGV caused by the problem, #390 may be closed anyway.


jmle:
@ascheman could you sign all the commits?

PR #430, opened 2023-11-15
:bug: using golang for filecontent searching

shawn-hurley:
Working on increasing the performance inside podman VM's on Mac and Windows.

### DIFF
--- a/provider/internal/builtin/service_client.go
+++ b/provider/internal/builtin/service_client.go
@@ -1,14 +1,13 @@
 package builtin
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io/fs"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"regexp"
-	"strconv"
 	"strings"
 	"sync"
 
@@ -16,6 +15,7 @@ import (
 	"github.com/antchfx/xmlquery"
 	"github.com/antchfx/xpath"
 	"github.com/go-logr/logr"
+	"github.com/konveyor/analyzer-lsp/lsp/protocol"
 	"github.com/konveyor/analyzer-lsp/provider"
 	"github.com/konveyor/analyzer-lsp/tracing"
 	"go.lsp.dev/uri"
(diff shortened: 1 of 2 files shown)


shawn-hurley:
After this change:
 shurley@salamander  ~/repos/MTA/kyma_poc/data   main ±  RUNNER_IMAGE=f64c030abab688840f8e0db3ee4b16b5ca200d29c95e77f785c05eb6c3bc2921 ./analyze_coolstuff.sh
INFO[0000] running source code analysis                  args="--provider-settings=/opt/input/config/settings.json --rules=/opt/rulesets/ --output-file=/opt/output/output.yaml --context-lines=100 --dep-label-selector=(!konveyor.io/dep-source=open-source) --verbose=4 --label-selector=((konveyor.io/target=quarkus || konveyor.io/target=jakarta-ee || konveyor.io/target=jakarta-ee8+ || konveyor.io/target=jakarta-ee9+ || konveyor.io/target=cloud-readiness) && konveyor.io/source) || (discovery)" input=/Users/shurley/repos/MTA/kyma_poc/data/coolstuff-javaee log=/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee/analysis.log output=/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee volumes="{\"/Users/shurley/repos/MTA/kyma_poc/data/coolstuff-javaee\":\"/opt/input/source\",\"/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee\":\"/opt/output\",\"/var/folders/vt/5bfp7vyd1h79_7k5ygr0fttr0000gn/T/analyze-config-1528473597\":\"/opt/input/config\",\"/var/folders/vt/5bfp7vyd1h79_7k5ygr0fttr0000gn/T/m2-repo-597906230\":\"/root/.m2\"}"
INFO[0000] generating analysis log in file               file=/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee/analysis.log
INFO[0208] running dependency analysis                   args="--provider-settings=/opt/input/config/settings.json --rules=/opt/rulesets/ --output-file=/opt/output/output.yaml --context-lines=100 --dep-label-selector=(!konveyor.io/dep-source=open-source) --verbose=4 --label-selector=((konveyor.io/target=quarkus || konveyor.io/target=jakarta-ee || konveyor.io/target=jakarta-ee8+ || konveyor.io/target=jakarta-ee9+ || konveyor.io/target=cloud-readiness) && konveyor.io/source) || (discovery)" input=/Users/shurley/repos/MTA/kyma_poc/data/coolstuff-javaee log=/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee/dependency.log output=/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee
INFO[0208] generating dependency log in file             file=/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee/dependency.log
INFO[0262] generating static report                      args="/usr/local/bin/js-bundle-generator --analysis-output-list=/opt/output/output.yaml --deps-output-list=/opt/output/dependencies.yaml --output-path=/usr/local/static-report/output.js --application-name-list=coolstuff-javaee && cp -r /usr/local/static-report /opt/output" output=/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee
INFO[0263] Static report created. Access it at this URL:  URL="file:/Users/shurley/repos/MTA/kyma_poc/data/example_reports/coolstuff-javaee/static-report/index.html"

real    4m24.079s
user    0m0.114s
sys     0m0.546s

Before this change on Mac OS we were seeing around ~20 min

shawn-hurley:
@fabianvf maybe we should have done this from the beginning, my bad.

pranavgaikwad:
@shawn-hurley is there a concern of memory, when loading the entire file up?

provider/internal/builtin/service_client.go
```diff
@@ -284,3 +262,85 @@ func findFilesMatchingPattern(root, pattern string) ([]string, error) {
...
+		positions = append(positions, pos)
+	}
+
+	return positions, nil
+}
+
+func processFile(path string, regex *regexp.Regexp, positionsChan chan<- walkResult, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	content, err := os.ReadFile(path)
```
pranavgaikwad, line 308: Little concerned about memory here. I think we don't have limits on workers either. Should we have another check to switch `forceFullFileScan` if file size is beyond a certain value?

rule-example.yaml
```diff
@@ -184,7 +184,7 @@
   ruleID: filecontent-codesnip-test
   when:
     builtin.filecontent:
-      pattern: "^FROM.*openjdk-11.*"
+      pattern: ".*openjdk-11.*"
```
pranavgaikwad, line 187: Why not keep ^?

shawn-hurley: op my mistake, will fix

provider/internal/builtin/service_client.go
shawn-hurley: is the idea that going line by line and bringing it into the buffer will reduce the memory. 

The problem could arise if we have a jar file that is GB big, but even then, I don't think that going line by line is going to reduce the size of memory usage because do those compacted files have line breaks?

pranavgaikwad: @shawn-hurley I am suggesting that, beyond a certain size, we shouldn't `ReadFile` and only do the Scan(). I wasn't really worried about binaries, but in general, big files (for instance, a db seed). 

If the line break isn't present within first 64K, Scanner() will fail I think. So that's an edge case to that I didn't think about earlier.

pranavgaikwad: If there's a way to get rid of `ReadFile` and only do scan in chunks of pre-determined size would be even better. That will affect the accuracy unless we have a window mechanism.

shawn-hurley: I think I am suggesting that we live with this for now; if we see an issue in testing all the stuff, then let's come back. I can add a TODO that this may need to be changed.

Thoughts?

pranavgaikwad: I am fine with addressing this as a TODO. Thank you @shawn-hurley this looks great

pranavgaikwad (approved):

provider/internal/builtin/service_client.go
shawn-hurley: Captured concern in Issue: https://github.com/konveyor/analyzer-lsp/issues/433

PR #431, opened 2023-11-15
:bug: Implement sorting on structs when marshaling

### LINKED ISSUE #418: [BUG] marshalling of output is non-deterministic

### Konveyor version

e73e5df290d7c74aa1e58b60c1c7356252822984

### Priority

Major

### Current Behavior

When running integration tests (either locally, or on Github), the output of the analyzer/dep-analyzer sometimes has a different ordering. In this case, [tests based on `yq` and `diff`](https://github.com/konveyor/analyzer-lsp/blob/d2956153fc84e43fb05a8bec96f53061649e698a/.github/workflows/demo-testing.yml#L30) fail, though there is no semantic difference in the results.

### Expected Behavior

Output should be stable and preserve ordering when executed on the same input.

### How Reproducible

Sometimes

### Steps To Reproduce

export PODMAN_BIN=docker # or 'podman' of course ...
... # Build images like in the mentioned Github Action
${PODMAN_BIN} run -v $(pwd)/demo-output.yaml:/analyzer-lsp/output.yaml:Z localhost/testing:latest
diff \
            <(sort <(sed 's/incidents\.[0-9]\+/incidents\.x/g' <(yq -P 'sort_keys(..)' -o=props <(git show HEAD:demo-output.yaml)))) \
            <(sort <(sed 's/incidents\.[0-9]\+/incidents\.x/g' <(yq -P 'sort_keys(..)' -o=props <(cat demo-output.yaml))))

### Environment

_No response_

### Anything else?

* The problem was [already discussed in Slack](https://kubernetes.slack.com/archives/C04QZJFQ0UA/p1699894451823839) (thanks to @shawn-hurley and @JonahSussman)
* When marshalling is done in a deterministic way, comparison of results could be simplified

JonahSussman:
Fixes #418 

`demo-output.yaml`'s diff looks really ugly, but all this PR does is shuffle some stuff around. 

With this, we may not the `sed` and `yq` stuff in this workflow anymore:

https://github.com/konveyor/analyzer-lsp/blob/740142ea98dbd39b22a63e2da7bfa2374b86be29/.github/workflows/demo-testing.yml#L20-L25

### DIFF
--- a/output/v1/konveyor/violations.go
+++ b/output/v1/konveyor/violations.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"go.lsp.dev/uri"
+	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


output/v1/konveyor/violations.go
```diff
@@ -104,6 +150,46 @@ type Dep struct {
 	FileURIPrefix      string                 `json:"prefix,omitempty" yaml:"prefix,omitempty"`
 }
 
+func (d Dep) MarshalYAML() (interface{}, error) {
+	sort.Strings(d.Labels)
+
+	return d, nil
+}
+
+func (d Dep) cmpLess(other Dep) bool {
```
shawn-hurley, line 159: What is this used for?

shawn-hurley (changes requested):
I'm just thinking out loud here, but what do we think about doing this for JSON while we do this?

I don't think encoding/json does sorting by default so it may be worth it.

output/v1/konveyor/violations.go
```diff
@@ -73,6 +83,42 @@ type Violation struct {
...
+		}
+
+		return false
+	})
+	sort.SliceStable(v.Links, func(i, j int) bool {
+		if v.Links[i].URL != v.Links[j].URL {
+			return v.Links[i].URL < v.Links[j].URL
+		}
+
+		if v.Links[i].Title != v.Links[j].Title {
```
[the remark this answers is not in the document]
JonahSussman: I agree, changed!

output/v1/konveyor/violations.go
JonahSussman: I accidentally left it in before committing. I was doing some exploration on both JSON and YAML.

shawn-hurley (approved):
Looks good to me

PR #437, opened 2023-11-20
Analyze all (yet known) formats of Maven dependency graph elements

ascheman:
Analyze all (yet known) formats of Maven dependency graph elements.
Maven dependencies may have more or less than five elements when split up at the colon.

Contributing to #392 (by resolving yet known remaining issues).


jmle (approved):
Looks good, just a small question: is the GoDoc for `parseDepString` still valid? Might be worth updating it otherwise.

PR #441, opened 2023-11-28
:sparkles: LFX: Extend use-case of detecting deprecated Kubernetes API usage

Parthiba-Hazra:
Added 'yq' provider.
Added rules to detect use of deprecated APIs in k8s yaml files.


Parthiba-Hazra:
@JonahSussman @eemcmullan still need to add the test for those new function. Will try to complete that in few days.

Dockerfile
```diff
@@ -27,9 +30,12 @@ RUN python3 -m pip install python-lsp-server
 
 COPY --from=jaeger-builder /go/bin/all-in-one-linux /usr/bin/
 
+COPY --from=yq-builder /usr/bin/yq /usr/bin/yq
```
JonahSussman, line 33: Thinking out loud here, but I think there was talk of doing some dependency management for external providers a while back. It's fine for now, but this method could bloat the Dockerfile when adding more providers.

cmd/analyzer/main.go
```diff
@@ -217,7 +217,7 @@ func main() {
 	if err != nil {
 		log.Error(err, "error writing output file", "file", outputViolations)
 		os.Exit(1) // Treat the error as a fatal error
-	}	
+	}
```
JonahSussman, line 220: This extra space has been bugging me for a while now, thanks! :joy:

external-providers/yq-external-provider/pkg/generic/dependency.go
```diff
@@ -0,0 +1,39 @@
+package generic
```
JonahSussman, line 1: I would change the package name from `generic` to something like `yq_provider` to avoid conflating this with the generic-external-provider

external-providers/yq-external-provider/pkg/generic/dependency.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+
+	"github.com/konveyor/analyzer-lsp/provider"
+	"go.lsp.dev/uri"
+)
+
+func (g *genericServiceClient) GetDependencies(ctx context.Context) (map[uri.URI][]*provider.Dep, error) {
```
JonahSussman, line 13: Same as above, change instances of `genericServiceClient` to a name that's related to the provider itself.

external-providers/yq-external-provider/pkg/generic/dependency.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	m := map[uri.URI][]*provider.Dep{}
+	err = json.Unmarshal([]byte(data), &m)
+	if err != nil {
+		return nil, err
+	}
+	return m, err
+}
+
+func (p *genericServiceClient) GetDependenciesDAG(ctx context.Context) (map[uri.URI][]provider.DepDAGItem, error) {
```
JonahSussman, line 37: Same as above

external-providers/yq-external-provider/pkg/generic/provider.go
```diff
@@ -0,0 +1,125 @@
+package generic
```
JonahSussman, line 1: Same as above

external-providers/yq-external-provider/pkg/generic/provider.go
```diff
@@ -0,0 +1,125 @@
...
+			TemplateContext: openapi3.SchemaRef{},
+		},
+		{
+			Name:            "k8sResourceMatched",
+			TemplateContext: openapi3.SchemaRef{},
+		},
+	}
+}
+
+type genericCondition struct {
```
JonahSussman, line 43: Same as above

external-providers/yq-external-provider/pkg/generic/service_client.go
```diff
@@ -0,0 +1,317 @@
...
+	"github.com/go-logr/logr"
+	"github.com/konveyor/analyzer-lsp/jsonrpc2"
+	"github.com/konveyor/analyzer-lsp/lsp/protocol"
+	"github.com/konveyor/analyzer-lsp/provider"
+	"go.lsp.dev/uri"
+	"golang.org/x/mod/semver"
+	"gopkg.in/yaml.v2"
+)
+
+type genericServiceClient struct {
```
JonahSussman, line 28: Same as above

external-providers/yq-external-provider/pkg/generic/service_client.go
```diff
@@ -0,0 +1,317 @@
...
+
+var _ provider.ServiceClient = &genericServiceClient{}
+
+func (p *genericServiceClient) Stop() {
+	p.cancelFunc()
+	p.cmd.Wait()
+}
+
+func (p *genericServiceClient) Evaluate(ctx context.Context, cap string, conditionInfo []byte) (provider.ProviderEvaluateResponse, error) {
+	var cond genericCondition
```
JonahSussman, line 56: Maybe I'm not quite understanding, but does this support both "referenced" and "k8sResourceMatched" like stated in the `Capabilities` function? Looks like only the latter is supported here, which is fine. May need to update the `Capabilities` function

external-providers/yq-external-provider/pkg/generic/tables.go
JonahSussman, line 1: Unsure if this file is needed for the yq provider

k8s-rule.yaml
JonahSussman, line 1: Unsure if this is where we want to put these k8s rules. Perhaps we should add it to the https://github.com/konveyor/rulesets/ repo instead? WDYT @shawn-hurley

JonahSussman (changes requested):
Looks great, thanks for submitting a contribution! Just a few changes

external-providers/yq-external-provider/pkg/generic/service_client.go
Parthiba-Hazra: I think just need to keep k8sResourceMatched Capabilities for yq provider

external-providers/yq-external-provider/pkg/yq_provider/provider.go
```diff
@@ -0,0 +1,116 @@
...
+	if c.Proxy != nil {
+		// handle proxy settings
+		for k, v := range c.Proxy.ToEnvVars() {
+			os.Setenv(k, v)
+		}
+	}
+
+	lspServerPath, ok := c.ProviderSpecificConfig[provider.LspServerPathConfigKey].(string)
+	if !ok || lspServerPath == "" {
+		return nil, fmt.Errorf("invalid lspServerPath provided, unable to init go provider")
```
eemcmullan, line 78: this log mentions a go provider

external-providers/yq-external-provider/pkg/yq_provider/service_client.go
```diff
@@ -0,0 +1,317 @@
...
+
+type Release struct {
+	Name string `json:"name"`
+}
+
+const APIVERSION = "apiVersion"
+const KIND = "kind"
+const GITHUB_K8S_API_URL = "https://api.github.com/repos/kubernetes/kubernetes/releases"
+
+var default_k8s_version = "v1.28.4"
```
eemcmullan, line 46: we may want to think of a way to make this more dynamic

k8s-rule.yaml
```diff
@@ -0,0 +1,1034 @@
+- message: Deprecated/removed Kubernetes API version 'extensions/v1beta1' is used for 'Deployment'. Consider using 'apps/v1'.
```
eemcmullan, line 1: Could we maybe get these rules separated into rulesets by the api group? i.e. rules for apps, rules for networking, etc..

k8s-rule.yaml
```diff
@@ -0,0 +1,1034 @@
...
+  effort: 2
+  when:
+    yaml.k8sResourceMatched:
+        apiVersion: "extensions/v1beta1"
+        kind: "Deployment"
+        deprecatedIn: "v1.9.0"
+        removedIn: "v1.16.0"
+        replacementAPI: "apps/v1"
+- message: Deprecated/removed Kubernetes API version 'apps/v1beta2' is used for 'Deployment'. Consider using 'apps/v1'.
+  ruleID: k8s-deprecated-api-002
```
eemcmullan, line 14: If the above is done, we could then have more descriptive ruleIDs

eemcmullan (changes requested):

Dockerfile
shawn-hurley: It will for sure, I think we should ask @dymurray to prioritize this work for the next version.

k8s-rule.yaml
shawn-hurley, line 1: Can these rules be generated or did you do this by hand?

shawn-hurley (commented):
I think this makes a lot of sense; I wonder if we could make it an external provider in its own repo, though maybe in the konveyor-externsions and make a flag to build an image with it built in or something? 

Just wondering, if it is too much work right now I am cool with adding it and doing the messy work after :)

shawn-hurley:
@Parthiba-Hazra This is awesome work! Thank you so much for taking the time to work on this and contributing!

k8s-rule.yaml
Parthiba-Hazra: will look after it. need some time

Parthiba-Hazra: nah it's not generated

external-providers/yq-external-provider/pkg/yq_provider/service_client.go
Parthiba-Hazra: Should I add a config struct and then assign those values form a json file or yaml file?

k8s-rule.yaml
Parthiba-Hazra: I think we can find a way to get this automated.

shawn-hurley: Wow, that must have taken a long time, thank you. This is going to be a nice feature!

Parthiba-Hazra: trust me, that didn't take much time :sweat_smile:

shawn-hurley:
@Parthiba-Hazra can you rebase then I think we can merge

JonahSussman (commented):
LGTM!

JonahSussman (approved):

shawn-hurley (approved):

eemcmullan:
@Parthiba-Hazra Do you mind updating the demo testing output?

Parthiba-Hazra:
> @Parthiba-Hazra Do you mind updating the demo testing output?

yes sure,

k8s-rule.yaml
```diff
@@ -0,0 +1,1034 @@
+- message: Deprecated/removed Kubernetes API version 'extensions/v1beta1' is used for 'Deployment'. Consider using 'apps/v1'.
+  ruleID: k8s-deprecated-api-001
+  description: "Check for usage of deprecated Kubernetes API versions"
+  category: potential
+  effort: 2
+  when:
+    yaml.k8sResourceMatched:
+        apiVersion: "extensions/v1beta1"
+        kind: "Deployment"
+        deprecatedIn: "v1.9.0"
```
pranavgaikwad, line 10: What is the need for passing deprecated / removed versions as parameters here? IMO `k8sResourceMatched` should only care about matching on given fields in a resource and should have nothing to do with the deprecated versions etc. The rule author should use source / target labels to convey this information. Am I missing something?

pranavgaikwad (changes requested):

k8s-rule.yaml
Parthiba-Hazra: @pranavgaikwad yes in your approach we might don't need those parameters but during the project implementation we thought if it's possible to get the target version of k8s then users don't need to pick rules manually they can use the full rule-set for evaluation and application should create the violation upon the actual deprecated or removed api-versions according to target version .

shawn-hurley: @pranavgaikwad Given that we have #466, I am wondering if we can work on that issue as a follow-up and merge this as is because main is tracking 0.4.0 now?

pranavgaikwad (approved):

Issue #358, opened 2023-10-05
[BUG] Files in the target folder are being analyzed


jmle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.3.0-alpha.5

### Priority

Major

### Current Behavior

Duplicate entries are appearing because some files inside the `target` folder are being analyzed.

### Expected Behavior

When analyzing a source code project with "source only", the `target` folder appearing as a result of running maven compilation should not be included in the analysis.

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Given the environment, analyze the application using "Source + dependencies".
2. Look for the issue "Replace the Java EE persistence namespace, schemaLocation and version with the Jakarta equivalent", with rule ID `javaee-to-jakarta-namespaces-00002`.
3. Click on "View affected files" and see how two files within a `target` folder are included:
![image](https://github.com/konveyor/analyzer-lsp/assets/4527008/47c722a5-1eb0-4e4f-bae4-83522581dcdf)

### Environment

```markdown
* Fedora 38
* minikube version: v1.31.2
* Konveyor 0.3.0-alpha.5
* [sample.daytrader7 app (master branch)](https://github.com/WASdev/sample.daytrader7)
* Analysis configuration:
** Source code + dependencies
** eap8, quarkus, openjdk17, azure, cloud-readiness
** Application and internal dependencies only
** No custom rules
** No additional advanced options
```

jmle:
@shawn-hurley this is mainly producing duplicate incidents, so I think it's not a big problem.

jwmatthews:
I hit this with https://github.com/deewhyweb/eap-coolstore-monolith/tree/main

What surprised me is I didn't intentionally compile the application, so was confused when I saw references to a 'target' directory.  I now assume that during analysis the target directory was created and hence analyzed.

The real thing that threw me off with this is I was programmatically grabbing the files from Git that the analyzer flagged as an issue, and when I hit the 'target/*' files they were not in Git so I threw an exception in my code.  For now I'll just work around this


```

javaee-to-jakarta-namespaces-00002:
      description: |-
        Replace the Java EE persistence namespace, schemaLocation and version with the Jakarta equivalent
        Replace `http://xmlns.jcp.org/xml/ns/persistence` with `https://jakarta.ee/xml/ns/persistence` and change the schema version number
      category: mandatory
      labels:
      - konveyor.io/target=jakarta-ee9+
[...]

PR #446, opened 2023-12-06
:bug: Skip "target" folder in file analysis

aufi:
Updating filecontent condition to exclude files in "target" subdirectory. This condition already used `grep`, so adding `--exclude-dir` option there.

Fixes https://github.com/konveyor/analyzer-lsp/issues/358

### DIFF
--- a/provider/internal/builtin/service_client.go
+++ b/provider/internal/builtin/service_client.go
@@ -85,7 +85,7 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
 			return response, fmt.Errorf("could not parse provided regex pattern as string: %v", conditionInfo)
 		}
 		var outputBytes []byte
-		grep := exec.Command("grep", "-o", "-n", "-R", "-P", c.Pattern, p.config.Location)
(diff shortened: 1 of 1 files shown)


aufi:
@jmle I'm not sure such simple fix is appropriate here, thinking about listing `target` directories first and check they are not in git, so were likely generated recently and should be skipped. (and if there is no git repo, skip target automatically or do some changed timestamp check?)

jmle:
Hey @aufi, sorry I somehow missed this one last month. The main problem I see with this approach is that the `filecontent` condition is in the internal provider, and therefore not exclusive to Java, but the `target` problem will only happen there. I'm also wondering if this problem is exclusive to the `filecontent` condition, or could also appear for instance with the `javaclass` conditions.

shawn-hurley:
@aufi Hey, 

So I am going to close this, it was fixed in a different way in PR #534 

Please let me know if you think that this is not correct or if this fixes something else and we can reopen

aufi:
#534 covers this issue, so no problem with closing this PR, thanks Shawn!

PR #448, opened 2023-12-08
:bug: Adding internal provider to provide the package for a given match.

shawn-hurley:


shawn-hurley:
After some ideas around trying to parse out the package form URI, I decided that with the introduction of the multi-module pom stuff the best course of action here is to go to the source (the file) and get the package from that. 

If we start hitting more file i/o issues we should get the code snip here and cache it

fabianvf (approved):
LGTM

PR #450, opened 2023-12-14
:sparkles: Changes for using message instead of description

jmle:
This will make the tests fail, as the output changes with a new field (message).


jmle:
@shawn-hurley I did this for Kantra, since the static report there works slightly different, but with the refactor we need to do this might not be relevant anymore.

pranavgaikwad:
Let's re-open this when we start working on this again.

PR #461, opened 2024-01-08
:bug: Use window of correct size in presence of newlines to grab content from xml files

pranavgaikwad:
We are stripping out empty lines when doing the search, however, in the window that we load into memory we still have to load these empty lines to be able to correctly search content. We only have to get rid of newlines from beginning and the end.

### DIFF
--- a/provider/internal/builtin/service_client.go
+++ b/provider/internal/builtin/service_client.go
@@ -333,7 +333,7 @@ func (b *builtinServiceClient) getLocation(ctx context.Context, path, content st
 	if len(lines) > 0 && lines[len(lines)-1] == "" {
 		lines = lines[:len(lines)-1]
 	}
-	if len(lines) < 1 {
+	if len(lines) < 1 || strings.Join(lines, "") == "" {
 		return location, fmt.Errorf("unable to get code location, empty content")
 	}
 	pattern := fmt.Sprintf(".*?%s", strings.Join(lines, ".*?"))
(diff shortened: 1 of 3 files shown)


provider/internal/builtin/service_client.go
```diff
@@ -324,15 +324,19 @@ func (b *builtinServiceClient) getLocation(ctx context.Context, path, content st
 		line := strings.Trim(part, " ")
 		line = strings.ReplaceAll(line, "\t", "")
 		line = regexp.QuoteMeta(line)
-		if line != "" {
-			lines = append(lines, line)
-		}
+		lines = append(lines, line)
+	}
+	// remove leading and trailing empty lines
+	if len(lines) > 0 && lines[0] == "" {
```
shawn-hurley, line 330: Is there a time/place where you can have content that is 5 empty lines?

pranavgaikwad: @shawn-hurley content comes from innerText returned by the XML search, and it depends on the xml query, so theoritically it's possible if someone is intentionally / mistakenly looking for empty lines. in that case, we will end up showing empty lines as code snip

shawn-hurley: What if as we loop, as long as one thing has content, then we use it, if all don't then we skip it?

provider/internal/builtin/service_client.go
```diff
@@ -324,15 +324,19 @@ func (b *builtinServiceClient) getLocation(ctx context.Context, path, content st
...
+		lines = lines[1:]
 	}
-	if len(lines) < 1 {
-		return location, fmt.Errorf("unable to get code location, no-op content")
+	if len(lines) > 0 && lines[len(lines)-1] == "" {
+		lines = lines[:len(lines)-1]
+	}
+	if len(lines) < 1 || strings.Join(lines, "") == "" {
```
pranavgaikwad, line 336: @shawn-hurley this check should take care of the empty case

shawn-hurley: +1

shawn-hurley (approved):
Looks good to me

PR #464, opened 2024-01-10
:ghost: adding documentation for chaining conditions

### LINKED ISSUE #463: Document usage of from / as keywords when chaining conditions in a rule

shawn-hurley:
Inside and/or condition blocks.

fixes #463

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -116,7 +116,7 @@ There are three types of conditions - _and_, _or_ and _provider_. While the _pro
 
 #### Provider Condition
 
-The analyzer engine enables multi-language source code analysis via something called as “providers”. A "provider" knows how to analyse the source code of a technology. It publishes what it can do with the source code in terms of "capabilities".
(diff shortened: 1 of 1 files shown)


pranavgaikwad (approved):

docs/rules.md
```diff
@@ -283,9 +283,54 @@ when:
...
+You can also chain the variables from one condition to be used as input in another condition in a _or_ and _and_ block of conditions
+
+Example:
+
+```yaml 
+when:
+ or:
+  - builtin.xml:
+      xpath: "//dependencies/dependency"
+      filepaths: "{{poms.filepaths}}"
```
jmle, line 297: @shawn-hurley how can the user know what variables are available to get here? In this case it's `filepaths`, but what about the case of java conditions for instance?

shawn-hurley: This should be published by the provider int he capabilities for a given condition. 

With that said, I don't think we have a good way of exposing this data at the moment.

PR #473, opened 2024-01-23
:ghost: Improve running as non-root

jmontleon:
With this change we can at least run as non-root.

I would also proprose moving binaries that need to be accessible to all users be installed in /usr/local/bin. But as of right now it seems we have configs (or something) hard coded to this location.
https://github.com/konveyor/java-analyzer-bundle/blob/main/Dockerfile#L41

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html


shawn-hurley (approved):

shawn-hurley:
follow up issue created in #474

PR #499, opened 2024-02-07
:bug: respect match-all label values

pranavgaikwad:


shawn-hurley (changes requested):
This behavior feels wrong to me, if someone sets a source they want, we should still limit to that source. if they want any sources, then they should have a selector that allows for it `konveyor.io/source || konveyor.io/source=A` will give you any rules that have the source label with no value, as well as rules with source value at A. 

Unless I am missing something?

shawn-hurley (approved):
I do think that with this change, we need to make it clear how the empty label (label w/ no value. i.e konveyor.io/source) does in the docs. 

I don't have a good other option, the use case makes sense.

PR #505, opened 2024-02-15
:bug: Write errors to stderr

### LINKED ISSUE #479: [BUG] send errors to stderr

### Konveyor version

latest

### Priority

Major

### Current Behavior

errors are logged like so https://github.com/konveyor/analyzer-lsp/blob/main/cmd/analyzer/main.go#L100 

### Expected Behavior

_No response_

### How Reproducible

Always (Default)

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

In kantra, we have a log file for analyzer logs. There is an issue then with a user hitting an error and it not being apparent. We should send these errors to stderr as well.

eemcmullan:
Closes #479

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -76,19 +76,22 @@ func main() {
 	// This will globally prevent the yaml library from auto-wrapping lines at 80 characters
 	yaml.FutureLineWrap()
 
-	err := validateFlags()
-	if err != nil {
-		fmt.Printf("%v\n", err)
-		os.Exit(1)
-	}
-
 	logrusLog := logrus.New()
 	logrusLog.SetOutput(os.Stdout)
 	logrusLog.SetFormatter(&logrus.TextFormatter{})
 	// need to do research on mapping in logrusr to level here TODO
 	logrusLog.SetLevel(logrus.Level(logLevel))
-
 	log := logrusr.New(logrusLog)
+
+	logrusErrLog := logrus.New()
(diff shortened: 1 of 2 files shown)


shawn-hurley:
is there a way to not remove the errors from the other log/logfile as well?

shawn-hurley (approved):

pranavgaikwad:
@eemcmullan needs rebase

aufi (approved):
LGTM

cmd/analyzer/main.go
```diff
@@ -219,6 +230,7 @@ func main() {
 	err = os.WriteFile(outputViolations, b, 0644)
 	if err != nil {
 		log.Error(err, "error writing output file", "file", outputViolations)
+		errLog.Error(err, "error writing output file", "file", outputViolations)
```
pranavgaikwad, line 233: @eemcmullan is this printing error twice on the console? if yes, why not only keep stderr

pranavgaikwad (changes requested):

cmd/analyzer/main.go
shawn-hurley: @pranavgaikwad I asked that we make sure that the error still went to the log file and not just to Stderr. if this is printing twice @eemcmullan I think you probably need to look for a MultiWriter for the file and stdout when a file is set up: https://github.com/Sirupsen/logrus/issues/230#issuecomment-381639138.

pranavgaikwad (approved):

PR #515, opened 2024-02-21
:bug: Fix duplicated conditions when AS exists

jmle:
The parser was generating a duplicate entry in the list of conditions.


shawn-hurley (approved):
Thank you for making the system safer and fixing this bug!

aufi:
It looks I worked on fixing similar issue in MTA-2006 (patch below), but fixing it directly in rule parsing instead of evaluating seems to be much better.

```
diff --git a/engine/conditions.go b/engine/conditions.go
index 59936c8..facd4e4 100644
--- a/engine/conditions.go
+++ b/engine/conditions.go
@@ -277,10 +277,11 @@ func incidentsToFilepaths(incident []IncidentContext) []string {
 
 func sortConditionEntries(entries []ConditionEntry) []ConditionEntry {
        sorted := []ConditionEntry{}
-       for _, e := range entries {
+       for i, e := range entries {
                // entries without chaining or that begin a chain come first
                if e.From == "" {
-                       sorted = append(sorted, gatherChain(e, entries)...)
+                       subEntries := append(entries[:i], entries[i+1:]...)
+                       sorted = append(sorted, gatherChain(e, subEntries)...)
                }
        }
 
@@ -289,9 +290,14 @@ func sortConditionEntries(entries []ConditionEntry) []ConditionEntry {
 
 func gatherChain(start ConditionEntry, entries []ConditionEntry) []ConditionEntry {
        chain := []ConditionEntry{start}
-       for _, d := range entries {
-               if start.As == d.From && start.As != "" {
-                       chain = append(chain, gatherChain(d, entries)...)
+       haveNext := false
+       for i, d := range entries {
+               if start.As == d.From && start.As != "" && !haveNext {
+                       subEntries := append(entries[:i], entries[i+1:]...)
+                       if d.From != start.From && d.As != start.As {
+                               chain = append(chain, gatherChain(d, subEntries)...)
+                               haveNext = true
+                       }
                }
        }
        return chain

```

PR #516, opened 2024-02-22
:sparkles: Adding ability to output spec data

shawn-hurley:
Adding the ability for the spec of the given components to be rendered in swagger open API. 

TODO:
- [x] Add ability to generate condition and output capabilities on server startup
- [x] Add the rest of the condition fields for the shared condition (From, As, ignore, not, etc).
- [x] Validate that the output generally works against a rule-example.yaml file for specifically builtin.filecontent
- [x] Update providers to all correctly generate the openapi
- [x] Need to figure out the "output" and how to display that correctly.


cmd/analyzer/main.go
```diff
@@ -242,3 +265,121 @@ func validateFlags() error {
 
 	return nil
 }
+
+func createOpenAPISchema(providers map[string]provider.InternalProviderClient, log logr.Logger) openapi3.Spec {
```
pranavgaikwad, line 269: this is awesome

shawn-hurley:
Given the nature of the swagger UI, we would need to define a rest API like a contract. For our purpose, this is probably not what we want right now, skipping the display.

parser/open_api.go
```diff
@@ -0,0 +1,156 @@
...
+					Schema: &openapi3.Schema{
+						Type: &arraytype,
+						Items: &openapi3.SchemaOrRef{
+							Schema: &openapi3.Schema{
+								Type: &stringtype,
+							},
+						},
+					},
+				},
+				// TODO: here we need to add each capbability and/or
```
shawn-hurley, line 99: Note update comment here to that we `update the when` in the create in main

provider/internal/java/dependency.go
```diff
@@ -134,9 +134,7 @@ func (p *javaServiceClient) GetDependenciesFallback(ctx context.Context, locatio
 	p.log.V(10).Info("Analyzing POM",
 		"POM", fmt.Sprintf("%s:%s:%s", pomCoordinate(pom.GroupID), pomCoordinate(pom.ArtifactID), pomCoordinate(pom.Version)),
 		"error", err)
-	if err != nil {
```
shawn-hurley, line 137: Note to reviewer: this is because this err is checked above this is a redundant call

parser/open_api.go
```diff
@@ -0,0 +1,156 @@
+package parser
+
+import "github.com/swaggest/openapi-go/openapi3"
+
+var (
+	stringtype openapi3.SchemaType = openapi3.SchemaTypeString
```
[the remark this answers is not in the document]
shawn-hurley: I think I use provider here, lets remove this and use provider's

pranavgaikwad (approved):
This is great, thank you

PR #525, opened 2024-03-07
:bug: stop building the project when we don't need anything else

shawn-hurley:


shawn-hurley:
@jmle @pranavgaikwad It looks like this change is a nice compromise for #358. We won't create a target directory anymore (as seen in the e2e tests), BUT if one exists, then it will be searched AFAICT. WDYT?

pranavgaikwad:
@shawn-hurley that sounds good to me. I noticed that in our e2e tests, we have an incident from target directory as well. We need to remove this https://github.com/konveyor/go-konveyor-tests/blob/c5e0436cd5662025e89a090d5cd88adfd30112d3/analysis/tc_tackle_testapp_public.go#L38-L43

shawn-hurley:
Yep, I am working on that PR now

Issue #529, opened 2024-03-08
[Bug] Dependency CLI doesn't need decompilation of every file


pranavgaikwad:
Right now, since we share the _Init()_ function of a provider between analyzer and dep CLI commands, the java provider ends up decompiling things twice as decomp happens in _Init()_. We can either combine both these commands OR have another _InitDep()_ function for providers that does a light weight init.

shawn-hurley:
Closed

PR #530, opened 2024-03-08
:bug: improve decompile performance

pranavgaikwad:
This improves decompilation performance significantly. Also adds more tracing spans so we get more visibility into decomp perf in future.

Before:

* Full analysis via kantra in source + deps mode
    _InProgress_

* Decompile performance (41 sec per file, 22451 sec total)

    Screenshot_20240308_133015

After:

* Full analysis via kantra in source + deps mode (improved by )
     ```
     real	10m21.651s
     user	0m1.978s
     sys	0m2.554s
     ```
     
* Decompile performance (3.1 sec on per file, 1683 sec total, improved by ~93%)

    Screenshot_20240308_133054

    
There's one more issue we need to fix. We are decompiling for dependency CLI too. That's because we use the same _Init()_ function for analyzer and dep cli. We either need to combine these two commands, or have an alternate _InitDep()_ function interface on providers that does a light weight init of providers. Captured in https://github.com/konveyor/analyzer-lsp/issues/529

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/swaggest/openapi-go/openapi3"
+	"go.opentelemetry.io/otel/attribute"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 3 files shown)


aufi (approved):
LGTM

shawn-hurley (approved):
Looks good to me, I think the new spans will be really helpful too!

PR #531, opened 2024-03-11
:sparkles: adding ability to run dep command during analysis command

shawn-hurley:
Make it so that the analysis command can also, in parallel can do the dependency output if desired


pranavgaikwad (approved):

pranavgaikwad:
@shawn-hurley we need the addon fix before merging this

aufi (approved):

PR #532, opened 2024-03-12
:bug: Update indirect Dependency location

aufi:
Updating dependency incident location lookup for indirect dependencies to point to "parent" dependency location. E.g. for `maven-javax-to-jakarta-00002` rule.

Related to https://issues.redhat.com/browse/MTA-2006


aufi:
Demo worked well for me locally, looking on it and fixing unit tests (that are failing also locally).

PR #534, opened 2024-03-12
:sparkles: stop building the project when we don't need anything else

shawn-hurley:
There is no need to wait for the build for the init step to be complete.

API tests PR: 102


pranavgaikwad (approved):

shawn-hurley:
I was not planning on it, @dymurray @pranavgaikwad  is this something that we had needed for 0.3?

PR #543, opened 2024-03-19
:bug: Fix file URI prefix for JARs and do not move found JARs so dependency CLI can still find them

### LINKED ISSUE #544: [Bug] Dependency file URIs of JARs are causing them to not match dep label selector
We are not adding "file://" prefix to the filepaths of dependencies that we find in JARs. We need to correctly handle those paths for dep label selector to match. Also, we are moving original JARs to m2 repo we should be copying them so provider can still find them in their original place. Fixing this will fix dep-label-selector option correctly for JARs.

pranavgaikwad:
Fixes #544


aufi (approved):
LGTM

PR #548, opened 2024-03-22
:sparkles:  Don't use full file paths for violations

### LINKED ISSUE #322: Full file paths for dependencies are irrelevant for the user

eemcmullan:
Closes #322

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
@@ -444,17 +443,10 @@ func DependencyOutput(ctx context.Context, providers map[string]provider.Interna
 				continue
 			}
 			for u, ds := range deps {
-				var outputDepDag []konveyor.DepDAGItem
-				// trim m2 repo prefix for java deps
-				if name == "java" {
-					outputDepDag = trimDagItemPrefixes(ds)
-				} else {
-					outputDepDag = ds
-				}
 				depsTree = append(depsTree, konveyor.DepsTreeItem{
 					FileURI:      string(u),
 					Provider:     name,
-					Dependencies: outputDepDag,
+					Dependencies: ds,
 				})
 			}
 		} else {
@@ -464,17 +456,11 @@ func DependencyOutput(ctx context.Context, providers map[string]provider.Interna
 				continue
 			}
 			for u, ds := range deps {
-				var outputDeps []*konveyor.Dep
-				// trim m2 repo prefix for java deps
-				if name == "java" {
-					outputDeps = trimFlatDepPrefixes(ds)
-				} else {
-					outputDeps = ds
-				}
+				newDeps := ds
 				depsFlat = append(depsFlat, konveyor.DepsFlatItem{
 					Provider:     name,
 					FileURI:      string(u),
-					Dependencies: outputDeps,
+					Dependencies: newDeps,
 				})
 			}
 		}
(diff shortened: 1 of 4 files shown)


shawn-hurley (approved):
I think you have to fix the dependency test as well: 

https://github.com/konveyor/analyzer-lsp/actions/runs/8394658124/job/22992283193?pr=548#step:4:569

But other than that, this looks great and makes a ton of sense.

aufi (approved):
I agree with Shawn; need fix tests, otherwise looks great.

provider/internal/java/dependency.go
```diff
@@ -411,7 +411,7 @@ func (p *javaServiceClient) parseDepString(dep, localRepoPath, pomPath string) (
 	fp := resolveDepFilepath(&d, p, group, artifact, localRepoPath)
 
 	d.Labels = addDepLabels(p.depToLabels, d.Name)
-	d.FileURIPrefix = fmt.Sprintf("file://%v", filepath.Dir(fp))
+	d.FileURIPrefix = fmt.Sprintf("file://%v", filepath.Dir(strings.TrimPrefix(fp, localRepoPath)))
```
pranavgaikwad, line 414: I don't think we should be doing this,

pranavgaikwad (changes requested):

provider/internal/java/dependency.go
shawn-hurley: why?

pranavgaikwad: this will break the dependency filtering logic, either we fix it in both places or don't change the fileURIPrefix at all, because this is mainly just an internal field

shawn-hurley: @eemcmullan this could be causing the failures too then maybe?

eemcmullan: Think so. Probably best then to trim this prefix in https://github.com/konveyor/analyzer-lsp/blob/main/cmd/analyzer/main.go#L423

cmd/analyzer/main.go
```diff
@@ -437,10 +444,17 @@ func DependencyOutput(ctx context.Context, providers map[string]provider.Interna
 				continue
 			}
 			for u, ds := range deps {
+				var outputDepDag []konveyor.DepDAGItem
+				// trim m2 repo prefix for java deps
+				if name == "java" {
+					outputDepDag = trimDagItemPrefixes(ds)
```
pranavgaikwad, line 450: sorry one small request...I think hardcoding for java provider here is unnecessary. there is no need to change prefix fields on dependencies as it isn't output to users in any incidents.

pranavgaikwad (changes requested):
Sorry there's one small thing that concerns me, otherwise looking good. Thank you!

cmd/analyzer/main.go
eemcmullan: I think this is for the `--dep-output-file` for the user. Just checking for the java provider here as the trim() function it's calling is only used to trim the local m2 repo prefix. Wanted to avoid issue with other dep providers. However, if I'm off-track, please let me know.

pranavgaikwad: I think it is fine to leave the prefixes as-is in the output @shawn-hurley wdyt?

pranavgaikwad (approved):

PR #557, opened 2024-03-27
:bug: Update indirect Dependency location

aufi:
Updating dependency incident location lookup for indirect dependencies to point to "parent" dependency location. E.g. for maven-javax-to-jakarta-00002 rule.

Replaces https://github.com/konveyor/analyzer-lsp/pull/532

Related to https://issues.redhat.com/browse/MTA-2006


rule-example.yaml
```diff
@@ -241,4 +241,28 @@
       - beans.xml
       namespaces:
         b: http://xmlns.jcp.org/xml/ns/javaee
-      xpath: /b:beans
\ No newline at end of file
+      xpath: /b:beans
+- category: potential
+  customVariables: []
```
pranavgaikwad, line 246: nit: can we shrink this rule, just keep the required fields for brevity i.e. remove labels, links, customVariables?

demo-dep-output.yaml
```diff
@@ -361,6 +374,19 @@
     resolvedIdentifier: 7873092d39ef741575ca91378a6a21c388363ac8
     extras:
       artifactId: logback-core
+      baseDep:
+        name: ch.qos.logback.logback-classic
```
pranavgaikwad, line 378: nit: can we produce only the required fields of the baseDep

provider/provider.go
```diff
@@ -745,6 +745,15 @@ func (dc DependencyCondition) Evaluate(ctx context.Context, log logr.Logger, con
 			if err == nil {
 				incident.LineNumber = &location.StartPosition.Line
 				incident.CodeLocation = &location
+			} else if baseDep, ok := matchedDep.dep.Extras["baseDep"]; ok {
```
pranavgaikwad, line 748: since I looked into too many perf issues recently :)... maybe we can make only one call to GetLocation based on either direct or base dep?

pranavgaikwad (changes requested):
Thank you for the fix, overall looks good. Requesting minor changes

aufi:
Thanks for review @pranavgaikwad, will update!

pranavgaikwad (approved):
@aufi Thank you! LGTM

PR #560, opened 2024-04-02
:bug: fix demo output

pranavgaikwad:
This slipped through somehow after merging #351 I should have re-run CI before merging #557


aufi (approved):
LGTM

PR #566, opened 2024-04-04
:bug: Don't use full konveyor.Dep for dependency Extras

eemcmullan:


pranavgaikwad:
@aufi Not all complex fields are a problem only a certain fields that make grpc protobuf / json / yaml marshaling get confused on some nested fields. In this case, labels []string was being an issue

provider/provider.go
```diff
@@ -742,8 +743,18 @@ func (dc DependencyCondition) Evaluate(ctx context.Context, log logr.Logger, con
 			// this is a best-effort step and we don't want to block if resolver misbehaves
 			timeoutContext, cancelFunc := context.WithTimeout(context.Background(), time.Second*3)
 			if baseDep, ok := matchedDep.dep.Extras["baseDep"]; ok {
+				// convert base dep back to konveyor.Dep
+				konvDep := konveyor.Dep{}
+				depBytes, err := json.Marshal(baseDep)
+				if err != nil {
+					cancelFunc()
```
pranavgaikwad, line 750: you don't need to cancel the conext here, as its not being used yet, only cancel once in the end, also, it'd be nice to log the error and move on instead.

pranavgaikwad (changes requested):
requesting a minor change

aufi (approved):

pranavgaikwad (approved):

PR #579, opened 2024-04-17
:sparkles: Add dockerfiles for providers and run the demo using the external providers in a Pod

### LINKED ISSUE #470: Split Dockerfiles into seperate provider images

pranavgaikwad:
Fixes #470 

I have still kept the original base Dockerfile as-is for backwards compatibility. This is the first step in which we are chaging our demo tests to use the external providers. I will create a follow up issue to track the work around actually making the switch in the addon.

### DIFF
--- a/.github/workflows/demo-testing.yml
+++ b/.github/workflows/demo-testing.yml
@@ -86,7 +86,7 @@ jobs:
             elif .name == "yaml" then del(.binaryPath) | .address = "localhost:9995"
             else .
             end
-          )' provider_container_settings.json > provider_container_settings.json
+          )' provider_container_settings.json | tee provider_container_settings.json
 
           podman build -f demo.Dockerfile -t localhost/testing:latest
           podman run --entrypoint /usr/local/bin/konveyor-analyzer \
--- a/.github/workflows/demo-testing.yml
+++ b/.github/workflows/demo-testing.yml
@@ -10,38 +10,97 @@ jobs:
     steps:
       - name: Extract pull request number from inputs or PR description
         id: extract-info
+        env:
+          PULL_REQUEST_BODY: ${{ github.event.pull_request.body }}
         run: |
-          echo "${{ github.event.pull_request.body }}"
-          PULL_REQUEST_NUMBER=$(echo "${{ github.event.pull_request.body }}" | grep -oP 'Addon PR: \K\d+' || true)
-          echo "API_TESTS_REF=main" >> $GITHUB_OUTPUT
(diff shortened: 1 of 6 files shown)


.github/workflows/demo-testing.yml
```diff
@@ -10,38 +10,97 @@ jobs:
...
+
+          jq 'map(
+            if .name == "go" then del(.binaryPath) | .address = "localhost:9999"
+            elif .name == "nodejs" then del(.binaryPath) | .address = "localhost:9998"
+            elif .name == "python" then del(.binaryPath) | .address = "localhost:9997"
+            elif .name == "java" then del(.binaryPath) | .address = "localhost:9996"
+            elif .name == "yaml" then del(.binaryPath) | .address = "localhost:9995"
+            else .
+            end
+          )' provider_container_settings.json > provider_container_settings.json
```
djzager, line 89: Maybe want to see the container settings in the log?

```suggestion
          jq 'map(
            if .name == "go" then del(.binaryPath) | .address = "localhost:9999"
            elif .name == "nodejs" then del(.binaryPath) | .address = "localhost:9998"
            elif .name == "python" then del(.binaryPath) | .address = "localhost:9997"
            elif .name == "java" then del(.binaryPath) | .address = "localhost:9996"
            elif .name == "yaml" then del(.binaryPath) | .address = "localhost:9995"
            else .
            end
          )' provider_container_settings.json | tee provider_container_settings.json
```

At least I think that's close to what it would be 😎

.github/workflows/image-build.yaml
```diff
@@ -41,3 +57,18 @@ jobs:
     secrets:
       registry_username: ${{ secrets.QUAY_PUBLISH_ROBOT }}
       registry_password: ${{ secrets.QUAY_PUBLISH_TOKEN }}
+  
+  # generic provider requires go dep provider we built in matrix and needs to be sequential
+  generic-external-provider-build:
```
djzager, line 62: I think you want this to wait for [image-build to be successful](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs).

```suggestion
  generic-external-provider-build:
    needs: image-build
```

djzager (changes requested):
Nice work! Just a couple of things I think worth looking at before merge.

shawn-hurley:
Nothing looks wrong to me, this is really good! thank you!

djzager (approved):
lgtm. nice

PR #591, opened 2024-04-26
:sparkles: Gradle support

jmle:
(https://github.com/konveyor/analyzer-lsp/issues/598) Adds support for fetching dependencies with Gradle projects.

- This implementation assumes that there is a Gradle wrapper included in the project, otherwise it is not possible to know which Gradle version is needed.
- To maximize compatibility with most Gradle versions, JDK8 must be used. There are a number of ways to specify this on newer Gradle versions, but for older versions the only thing that seems to work is setting `JAVA_HOME`.
- Depends on https://github.com/konveyor/java-analyzer-bundle/pull/93

:exclamation: This PR does not provide support for fetching dependency sources yet.

:green_circle: Tested against https://github.com/andyjduncan/gradle-example and https://github.com/kissaten/gradle-multi-project-example


demo-dep-output.yaml
```diff
@@ -2332,4 +2463,4 @@
     labels:
     - konveyor.io/dep-source=open-source
     - konveyor.io/language=java
-    prefix: file:///root/.m2/repository/org/yaml/snakeyaml/1.30
+    prefix: file://org/yaml/snakeyaml/1.30
```
pranavgaikwad, line 2466: what happened to file URIs here?

external-providers/java-external-provider/pkg/java_external_provider/dependency.go
```diff
@@ -51,10 +67,41 @@ func (p *javaServiceClient) findPom() string {
...
+	if _, err := os.Stat(f); errors.Is(err, os.ErrNotExist) {
+		return ""
+	}
 	return f
 }
 
+func (p *javaServiceClient) findGradleBuild() string {
+	// TODO: naive?
+	if p.config.Location != "" {
+		f, err := filepath.Abs(filepath.Join(p.config.Location, "build.gradle"))
```
pranavgaikwad, line 79: is Abs() running os.Stat()?

external-providers/java-external-provider/pkg/java_external_provider/dependency.go
```diff
@@ -51,10 +67,41 @@ func (p *javaServiceClient) findPom() string {
...
+		if err != nil {
+			return ""
+		}
+		return f
+	}
+	return ""
+}
+
 func (p *javaServiceClient) GetDependencies(ctx context.Context) (map[uri.URI][]*provider.Dep, error) {
+	if p.getBuildTool() == gradle {
```
pranavgaikwad, line 89: is this redundant given that there's a switch in GetDepenenciesDAG()? I think we will need to move fallback call inside GetDepenciesDAG()

pranavgaikwad (changes requested):
small nits / questions...overall looks great!

external-providers/java-external-provider/pkg/java_external_provider/dependency.go
jmle: :+1:

demo-dep-output.yaml
jmle: This should be fixed now

jmle:
Superseded by https://github.com/konveyor/analyzer-lsp/pull/609

PR #592, opened 2024-04-29
:bug: Add openapi struct fields for dependency condition

### LINKED ISSUE #551: [Bug] Add openapi struct fields for dependency condition
using: https://github.com/swaggest/jsonschema-go?tab=readme-ov-file#field-tags

We should add the title, description, and pattern fields to the struct here:
https://github.com/konveyor/analyzer-lsp/blob/feature/use-open-api/provider/provider.go#L616-L624

You can use this as a starting point for the [semver](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) regular expression.

| Provider | Capability  | Fields     | Required | Description                                                   |
|----------|-------------|------------|----------|---------------------------------------------------------------|
|          | dependency  | name       | Yes      | Name of the dependency                                        |
|          |             | nameregex  | No       | Regex pattern to match the name                               |
|          |             | upperbound | No       | Match versions lower than or equal to                         |
|          |             | lowerbound | No       | Match versions greater than or equal to                       |

Vickysomtee:
fixes #551


pranavgaikwad (changes requested):

pranavgaikwad (approved):

Vickysomtee:
> @Vickysomtee can you rebase your PR with main?

@pranavgaikwad done ✅

PR #602, opened 2024-05-14
:sparkles: Adding optional TLS for provider communication

shawn-hurley:
Adding for all the in-tree providers the ability to use TLS authentication


djzager (approved):
Ack

eemcmullan (approved):

shawn-hurley:
@pranavgaikwad @eemcmullan @djzager I want to point out that this is not authentication, just that it uses certs to encrypt traffic over the wire. This is obviously strictly better but I don't think fully covers the full story.

PR #606, opened 2024-05-16
:sparkles: Path inclusion

pranavgaikwad:
Bundle PR: 94

### DIFF
--- a/provider/internal/builtin/service_client.go
+++ b/provider/internal/builtin/service_client.go
@@ -58,31 +58,25 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
 		if err != nil {
 			return response, fmt.Errorf("unable to find files using pattern `%s`: %v", c.Pattern, err)
 		}
-		matchingFiles = p.filterByIncludedPaths(matchingFiles)
-		if len(matchingFiles) != 0 {
-			response.Matched = true
-		}
 
 		response.TemplateContext = map[string]interface{}{"filepaths": matchingFiles}
 		for _, match := range matchingFiles {
-			if filepath.IsAbs(match) {
-				response.Incidents = append(response.Incidents, provider.IncidentContext{
-					FileURI: uri.File(match),
-				})
-				continue
-
+			absPath := match
+			if !filepath.IsAbs(match) {
+				absPath, err = filepath.Abs(match)
+				if err != nil {
+					p.log.V(5).Error(err, "failed to get absolute path to file", "path", match)
+					absPath = match
+				}
 			}
-			ab, err := filepath.Abs(match)
-			if err != nil {
-				//TODO: Probably want to log or something to let us know we can't get absolute path here.
-				fmt.Printf("\n%v\n\n", err)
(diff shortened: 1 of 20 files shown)


pranavgaikwad:
Finding a way to test this PR with bundle PR

provider/lib.go
```diff
@@ -142,3 +142,30 @@ func MultilineGrep(ctx context.Context, window int, path, pattern string) (int,
 
 	return -1, scanner.Err()
 }
+
+// GetIncludedPathsFromConfig returns validated includedPaths from provider settings
+// if allowFilePaths is not set, path to a file is converted into a path to its base dir
+func GetIncludedPathsFromConfig(i InitConfig, allowFilePaths bool) []string {
+	validatedPaths := []string{}
+	if includedPaths, ok := i.ProviderSpecificConfig[IncludedPathsConfigKey].([]interface{}); ok {
```
shawn-hurley, line 150: Do we want to error, or warn if it is not a slice?

Just to make debug of provider config easier?

provider/internal/builtin/service_client.go
```diff
@@ -268,18 +272,21 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
...
 				response.Matched = true
 				for _, node := range list {
-					ab, err := filepath.Abs(file)
+					absPath, err := filepath.Abs(file)
 					if err != nil {
-						ab = file
+						absPath = file
+					}
+					if !p.isFileIncluded(absPath) {
+						continue
```
shawn-hurley, line 280: Potentially add a high verbosity log here to note why it was removed.

pranavgaikwad:
I am investigating the failing test now, everything is working fine locally, looks like something small.

provider/lib.go
```diff
@@ -142,3 +142,30 @@ func MultilineGrep(ctx context.Context, window int, path, pattern string) (int,
 
 	return -1, scanner.Err()
 }
+
+// GetIncludedPathsFromConfig returns validated includedPaths from provider settings
+// if allowFilePaths is not set, path to a file is converted into a path to its base dir
```
shawn-hurley, line 147: Can you walk me through the use case for this option to be false?

provider/lib.go
pranavgaikwad: @shawn-hurley good point.

shawn-hurley (approved):
Overall, it looks really good to me; great job!

The only point I have is around the true/false option to use dir when getting a file path, just to understand when a user would want to set that to false.

Also, just a minor point but like 5 or 7 verbosity log when filtering out might be helpful in the future.

provider/lib.go
pranavgaikwad: @shawn-hurley this doesn't affect users, this is for providers...so with Java, we can only scope the path upto granularity of a package (as opposed to a class or file). So if user inputs `/util/File.java` we can only filter upto `/util`. I thought this would be potentially useful elsewhere, but I can push this specific logic into java provider too keeping this lib function generic

PR #608, opened 2024-05-21
:bug: fix rule parsing when sub-trees contain tests

pranavgaikwad:
This is needed because we now have "tests" folders in our rulesets. With current logic, if any sub-tree fails to load rulesets, we fail rule parsing altogether. We only want to fail on fatal errors.


aufi (approved):
LGTM

Issue #460, opened 2024-01-05
Add the ability to query on annotation information


jmle:
In windup, it is possible to further inspect an annotation in a condition [as in here](https://github.com/windup/windup-rulesets/blob/40d3c19512e8766afcbb209d7402826fe3fe84d6/rules/rules-reviewed/camel3/camel2/java-multiple-camelcontexts-per-application-not-supported.windup.xml#L17-L20) (see [the docs](https://access.redhat.com/documentation/en-us/migration_toolkit_for_applications/5.2/html-single/rules_development_guide/index#javaclass_child_elements)). It would be good to investigate if something similar can be achieved with the LS.

At the moment [we are not translating those attributes](https://github.com/konveyor/windup-shim/blob/main/pkg/conversion/convert.go#L438-L444) of the original rules.

PR #621, opened 2024-06-18
:bug: Temptative solution for basic annotation inspection

jmle:
Partially fixes https://github.com/konveyor/analyzer-lsp/issues/460

This PR would introduce a new field to the `java.referenced` condition. For instance, the following condition
```yaml
    java.referenced:
      pattern: org.home.Bean*
      location: CLASS
      annotated:
        pattern: org.home.BeanAnnotation
```
would match this:
```java
package org.home;

@org.home.BeanAnnotation
public class BeanClass {
  // ...
}
```
Possible values for annotated `location`s are `CLASS`, `METHOD` and `FIELD`.

:warning: When using `FIELD` in `java.referenced.location`, the user might expect to be able to put a type in the `pattern` field, as in "please look for any field of type `type` that has annotation `annotated.pattern`. That is not possible at the moment, since the LS only returns *the name* of the field.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/filter.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/filter.go
@@ -112,11 +112,11 @@ func (p *javaServiceClient) filterAnnotated(cond *javaCondition, symbols []proto
 			if kind == "class" {
 				// for classes, we need the fully qualified name (package + class name)
 				fqn := strings.Join([]string{pkg, name}, ".")
-				if pattern.Match([]byte(fqn)) && kind == strings.ToLower(cond.Referenced.Location) {
+				if pattern.Match([]byte(fqn)) {
 					incidents = append(incidents, incident)
 				}
 			} else if kind == "method" || kind == "field" {
-				if pattern.Match([]byte(name)) && kind == strings.ToLower(cond.Referenced.Location) {
+				if pattern.Match([]byte(name)) {
 					incidents = append(incidents, incident)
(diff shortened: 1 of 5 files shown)


external-providers/java-external-provider/pkg/java_external_provider/filter.go
```diff
@@ -91,6 +94,38 @@ func (p *javaServiceClient) filterDefault(symbols []protocol.WorkspaceSymbol) ([
...
+			if kind == "class" {
+				// for classes, we need the fully qualified name (package + class name)
+				fqn := strings.Join([]string{pkg, name}, ".")
+				if pattern.Match([]byte(fqn)) && kind == strings.ToLower(cond.Referenced.Location) {
+					incidents = append(incidents, incident)
+				}
+			} else if kind == "method" || kind == "field" {
+				if pattern.Match([]byte(name)) && kind == strings.ToLower(cond.Referenced.Location) {
+					incidents = append(incidents, incident)
+				}
```
abrugaro, line 121: The `kind == strings.ToLower(cond.Referenced.Location) ` condition in both ifs in lines 115 and 119 will be always true since the same condition exists in line 110

abrugaro (changes requested):

external-providers/java-external-provider/pkg/java_external_provider/filter.go
jmle: Good catch!

PR #635, opened 2024-07-02
:sparkles: Running builtin provider externally from a GRPC Provider

shawn-hurley:
Allowing for builtin to run, removing the need for shared access to the filesytem. Now only the provider will need access. 

Only implemented for java provider at the moment


shawn-hurley:
This is for 0.6

PR #657, opened 2024-07-15
:bug: handle progress messages from lang server

djzager:


external-providers/dotnet-external-provider/pkg/dotnet/provider.go
```diff
@@ -157,6 +157,9 @@ func (p *dotnetProvider) Init(ctx context.Context, log logr.Logger, config provi
 					},
 					// WorkspaceFolders: true,
 				},
+				Window: &protocol.WindowClientCapabilities{
+					WorkDoneProgress: false,
```
djzager, line 161: I don't think they care that we don't support it 😎

eemcmullan (approved):

Issue #467, opened 2024-01-11
[RFE] Ability to match on methods/functions/constructors with particular argument types


jmle:
The analyzer should have the ability to match on methods, functions or constructors that are passed a specific argument type, for instance:
```yaml
java.referenced:
  location: CONSTRUCTOR
  pattern: java.lang.String\(java.lang.StringBuilder\)
```
would match on:
```java
StringBuilder arg = new StringBuilder("Hello World!")'
String str = new String(arg);
```
but not on:
```java
String arg = "Hello World!";
String str = new String(arg);
```

We could either escape the parenthesis of the method call, as done in the yaml example above, or add an additional field, as in:
```yaml
java.referenced:
  location: CONSTRUCTOR
  arguments:
    - java.lang.String
  pattern: java.lang.String(*)
```

jmle:
There seem to be a few:
![image](https://github.com/konveyor/analyzer-lsp/assets/4527008/8c7372ff-4d25-4bb5-be9b-21cf3c3e7bb6)

I wouldn't say it's super important, but if it can be achieved we should definitely add it at some point.

shawn-hurley:
That makes sene to add it, I am not at all opposed, was trying to get a sense of how important it is

PR #659, opened 2024-07-16
:bug: Add annotation inspection support

jmle:
This PR adds support for inspecting annotations.

:green_circle: Things that can be done now:
- Check for annotations that are annotating specific `FIELD_DECLARATION`, `METHOD_DECLARATION` or `TYPE`
- Match against `METHOD_DECLARATION`

:red_circle: Things that cannot be done **yet**:
- Match `location: ANNOTATION` and inspect its values.
- Inspect on annotations on `location: CONSTRUCTOR`.

Depends on:
- https://github.com/konveyor/java-analyzer-bundle/pull/104

Fixes:
- https://github.com/konveyor/analyzer-lsp/issues/460
- https://github.com/konveyor/analyzer-lsp/issues/467

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -334,7 +334,7 @@ func (r *ruleEngine) runTaggingRules(ctx context.Context, infoRules []ruleMessag
 				rs.Errors[rule.RuleID] = err.Error()
 			}
 		} else if response.Matched && len(response.Incidents) > 0 {
-			r.logger.V(5).Info("info rule was matched", "ruleID", rule.RuleID)
+			r.logger.V(5).Info("rule was matched", "ruleID", rule.RuleID)
 			tags := map[string]bool{}
 			for _, tagString := range rule.Perform.Tag {
 				if strings.Contains(tagString, "{{") && strings.Contains(tagString, "}}") {
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -334,7 +334,7 @@ func (r *ruleEngine) runTaggingRules(ctx context.Context, infoRules []ruleMessag
 				rs.Errors[rule.RuleID] = err.Error()
 			}
 		} else if response.Matched && len(response.Incidents) > 0 {
-			r.logger.V(5).Info("rule was matched", "ruleID", rule.RuleID)
+			r.logger.V(5).Info("info rule was matched", "ruleID", rule.RuleID)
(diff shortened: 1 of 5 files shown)


engine/engine.go
```diff
@@ -334,7 +334,7 @@ func (r *ruleEngine) runTaggingRules(ctx context.Context, infoRules []ruleMessag
 				rs.Errors[rule.RuleID] = err.Error()
 			}
 		} else if response.Matched && len(response.Incidents) > 0 {
-			r.logger.V(5).Info("info rule was matched", "ruleID", rule.RuleID)
+			r.logger.V(5).Info("rule was matched", "ruleID", rule.RuleID)
```
shawn-hurley, line 337: why remove info here?

jmle: @shawn-hurley it sounded like a rule of type "info" was matched. It was confusing to me, as in windup we used to have INFO rules. But I can put it back if you want

external-providers/java-external-provider/pkg/java_external_provider/provider.go
```diff
@@ -296,6 +308,7 @@ func (p *javaProvider) Init(ctx context.Context, log logr.Logger, config provide
 		"-Djava.net.useSystemProxies=true",
 		"-configuration",
 		"./",
+		//"--jvm-arg=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:1044",
```
jmle, line 311: Uncommenting this line allows for the LS to be run in debugging mode. I find it useful, but can be removed if you guys thing it's not convenient.

engine/engine.go
shawn-hurley: can we updated to tagging rules, becuase they are seperate from the non-tagging rules

external-providers/java-external-provider/pkg/java_external_provider/provider.go
shawn-hurley: nope that is useful! assume that allows you to attach a debugger and set break points and stuff?

engine/engine.go
jmle: Then it probably makes sense to leave it as it was

pranavgaikwad (approved):

shawn-hurley (approved):

Issue #662, opened 2024-07-17
[BUG] Gradle analysis taking too long to complete


jmle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.5-beta2

### Priority

Critical

### Current Behavior

Currently, the analysis of the tackle-testapp-public in its gradle form takes too long to complete. This might be related to this bug opened in the hub:

https://github.com/konveyor/tackle2-hub/issues/667

### Expected Behavior

Gradle analysis shouldn't take too much longer than Maven analysis.

### How Reproducible

Always (Default)

### Steps To Reproduce

Analyze [tackle-testapp-public](https://github.com/konveyor/tackle-testapp-public) on its `main` branch (Maven) and then on its `gradle` branch (Gradle). Compare the time it takes for both to run.

jmle:
Can't reproduce anymore in the latest d/s

PR #663, opened 2024-07-17
:bug: Prevent gradle logic from decompressing jars all the time

jmle:
The gradle analysis was uncompressing jars every time because of a lack of a check for the existence of the already decompressed directory.

- Related to https://github.com/konveyor/tackle2-hub/issues/667
- Partly fixes https://github.com/konveyor/analyzer-lsp/issues/662

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/filter.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/filter.go
@@ -318,12 +318,14 @@ func (p *javaServiceClient) getURI(refURI string) (string, uri.URI, error) {
 		}
 		javaFileAbsolutePath = filepath.Join(filepath.Dir(sourcesFile), filepath.Dir(path), javaFileName)
 
-		cmd := exec.Command("jar", "xf", filepath.Base(sourcesFile))
-		cmd.Dir = filepath.Dir(sourcesFile)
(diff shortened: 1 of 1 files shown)


external-providers/java-external-provider/pkg/java_external_provider/filter.go
```diff
@@ -318,12 +318,14 @@ func (p *javaServiceClient) getURI(refURI string) (string, uri.URI, error) {
...
-		err = cmd.Run()
-		if err != nil {
-			fmt.Printf("\n java error%v", err)
-			return "", "", err
+		if _, err := os.Stat(filepath.Dir(javaFileAbsolutePath)); err != nil {
+			cmd := exec.Command("jar", "xf", filepath.Base(sourcesFile))
+			cmd.Dir = filepath.Dir(sourcesFile)
+			err = cmd.Run()
+			if err != nil {
+				fmt.Printf("\n java error%v", err)
```
pranavgaikwad, line 326: wondering if we should hard-fail here?

jmle: I would personally let the analysis continue, since it's only one JAR that failed decompressing

shawn-hurley (approved):

PR #665, opened 2024-07-18
:bug: unable to run binary path with maven output log

shawn-hurley:


eemcmullan:
@shawn-hurley  Is this something that'd be better to comment out and add a TODO until we understand more of the issue? To unblock using binaryPath for now

shawn-hurley:
I don't think we need this info it's a on of info and we handle the warnings and log when it errors and we fall back. I think I am ok with just removing it. Good thought though just me 2c if you feel strongly I can do it the other way I am not tied to either approach

eemcmullan:
No strong feelings. LGTM

eemcmullan (approved):

PR #668, opened 2024-07-22
:bug: use the location from the inital call

shawn-hurley:
I assume we used to find the class definition, and then had to find those references, with the updated call I can only assume we are getting the actual usage and the refenece is now finding the places that this is being called.


dymurray (approved):
VISIACK

pranavgaikwad (approved):
LGTM

PR #686, opened 2024-08-01
:bug: fix path of the downloaded maven artifact

pranavgaikwad:
This gets the path from maven output, falling back to old method, when it cannot.


shawn-hurley (approved):

shawn-hurley:
I assume that this needs to be cherry picked

PR #690, opened 2024-08-12
:sparkles: Windows Compatibility-Add PowerShell Script as an Alternative to Grep on Linux

kthatipally:
PR Summary
This pull request introduces a PowerShell-based alternative to the grep command, ensuring that the built-in File Searching feature is also compatible with Windows systems.

service_client.go: Added runOSSpecificGrepCommand to handle grep functionality for both Windows (using PowerShell) and Unix-based systems. This function checks for the OS platform and invokes powershell(windows) and grep(linux). Replaced direct grep command execution with runOSSpecificGrepCommand to support cross-platform compatibility.

service_client.go: Implemented parseGrepOutputForFileContent to parse the output from both PowerShell and grep. This enhancement not only refines the existing parsing mechanism for Linux but also accommodates Windows file path patterns and addresses issues with the split function when multiple ':' appear in the input.

WIP: Testing

### DIFF
--- a/provider/internal/builtin/service_client.go
+++ b/provider/internal/builtin/service_client.go
@@ -108,22 +108,11 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
 		}
 
 		var outputBytes []byte
-		if runtime.GOOS == "linux" {
-			if ok, paths := cond.ProviderContext.GetScopedFilepaths(); ok {
-				outputBytes, err = runOSSpecificGrepCommand(c.Pattern, paths...)
-			} else {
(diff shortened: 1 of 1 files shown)


shawn-hurley (approved):
This looks good assuming the power shell code works. 

Another option if you were interested in doing, is to create a new "grep_windows.go" and "grep_nix.go" that will compile grep_nix using build tags when on a nix system and windows when on windows. 

This is a decent resource to see what I mean: https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

I think that we can do this as a follow up though if we want

PR #695, opened 2024-08-28
:sparkles: Get deps fallback for .war/.ear files if no jars found

eemcmullan:


aufi (approved):
LGTM

jmle:
The only problem here is the case when there are multiple poms inside the WAR file, which is something that can happen. In that case, this solution would give a partial list of dependencies, since we are stopping after finding the first pom. Ideally we would find all poms and concat their respective deps

pranavgaikwad:
@jmle What happens in cases where the POMs have a parent - child relation? Does the fallback parser care if it cannot fully resolve the parent or children modules?

eemcmullan:
@jmle updated to get all poms

pranavgaikwad (approved):

PR #699, opened 2024-09-09
:sparkles: Update external provider modules

eemcmullan:
With the current external provider module names, we cannot import these providers into other projects, such as kantra. Updating the module names to the path inside of the main repo should fix this issue.


djzager (approved):

djzager:
Assuming we can get tests to pass, is there any reason to hold this PR @eemcmullan ? It would really be to my benefit to merge this.

aufi (approved):

djzager (approved):
🎉 Thank you @eemcmullan

PR #703, opened 2024-10-09
:sparkles: Run from /addon so we can run as non-root user

jmontleon:


external-providers/java-external-provider/Dockerfile
```diff
@@ -21,5 +21,5 @@ FROM quay.io/konveyor/jdtls-server-base
 COPY --from=builder /java-provider/java-external-provider /usr/local/bin/java-external-provider
 
 EXPOSE 14651
-
-ENTRYPOINT ["java-external-provider", "--port", "14651"]
\ No newline at end of file
+WORKDIR /addon
```
jortel, line 24: Would `/tmp` be a better (more agnostic) default?

jmontleon: @shawn-hurley @pranavgaikwad @fabianvf any opinions? `/tmp` sound good to me, but you all know far more about this project than I do.

dymurray (approved):

PR #705, opened 2024-10-11
:bug: Adding internal template context value for dependency

shawn-hurley:
This is to allow us to know when the `java.dependency` (or other dependency capabilities) are called. This will allow kai to know when the incident is targeting dependencies and will allow it to pass to the correct task_runner.


shawn-hurley:
Closing this, because we are going to solve it a different way.

PR #697, opened 2024-09-04
:sparkles: Print Maven output on error

jmle:
Fixes https://github.com/konveyor/analyzer-lsp/issues/696

PR #708, opened 2024-10-24
:sparkles: Log full Maven output on errors

aufi:
Java provider uses mvn command to fetch java binary or its dependencies. There was a check for success/error of those commands, but to identify a root cause and fix the mvn problem, its full output is required.

The maven output can be be exhaustive, but it is logged only in case of its failure.


aufi:
Looks to be a duplicate of https://github.com/konveyor/analyzer-lsp/pull/697

aufi:
Confirmed this can produce output to `java.log` files
image
(added same lines as error reporting to Info log before checking `err`, for testing purposes)

aufi:
Should be replaced with https://github.com/konveyor/analyzer-lsp/pull/709

Issue #718, opened 2024-11-11
[BUG] Condition chaining not working for `java.referenced` conditions


jmle:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

0.5.3

### Priority

Critical

### Current Behavior

A condition like the following:

```yaml
  when:
    or:
    - java.referenced:
        pattern: org.springframework.stereotype.Controller
        location: ANNOTATION
      as: class
      ignore: true
    - java.referenced:
        pattern: '* javax.xml.transform.Source'
        location: METHOD
      from: class
```
doesn't work because the chaining does not seem to work.

### Expected Behavior

Said condition should work. We should at least be sending the class file to the chain, and further filtering in the next condition, but that doesn't seem to be the case.

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Run the above rule against the following project:

[web-applications.zip](https://github.com/user-attachments/files/17701951/web-applications.zip)

2. One and only one issue should appear.

### Environment

```markdown

```

jmle:
Closed with https://github.com/konveyor/analyzer-lsp/pull/721

PR #721, opened 2024-11-14
:bug: Fix chaining in Java provider

jmle:
See https://github.com/konveyor/analyzer-lsp/issues/718

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -611,7 +611,7 @@ func templateCondition(condition []byte, ctx map[string]engine.ChainTemplate) ([
 	s := strings.ReplaceAll(string(condition), `'{{`, "{{")
 	s = strings.ReplaceAll(s, `}}'`, "}}")
 
-	s, err := mustache.Render(s, true, ctx)
+	s, err := mustache.RenderRaw(s, true, ctx)
 	if err != nil {
 		return nil, err
 	}
--- a/rule-example.yaml
+++ b/rule-example.yaml
@@ -402,3 +402,20 @@
         elements:
           - name: basePackages
             value: "io.konveyor.demo.ordermanagement.repository"
+- ruleID: java-chaining-01
+  category: mandatory
+  description: "There should only be one instance of this rule"
+  message: |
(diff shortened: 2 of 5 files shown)


rule-example.yaml
```diff
@@ -402,3 +402,20 @@
...
+    and:
+      - java.referenced:
+          pattern: java.lang.Override
+          location: ANNOTATION
+        as: class
+        ignore: true
+      - java.referenced:
+          pattern: org.springframework.web.WebApplicationInitializer
+          location: IMPLEMENTS_TYPE
+          filepaths: "{{class.Filepaths}}"
```
jmle, line 420: This is a PITA - Mustache wasn't being able to render `class.filepaths` because `Filepaths` lives inside a struct and is uppercased... We don't really want this, as it is something very implementation related

shawn-hurley: I wonder what happens if you and a Jain strict tag for filepaths that is lowercased in the struct

provider/provider.go
```diff
@@ -611,7 +611,7 @@ func templateCondition(condition []byte, ctx map[string]engine.ChainTemplate) ([
 	s := strings.ReplaceAll(string(condition), `'{{`, "{{")
 	s = strings.ReplaceAll(s, `}}'`, "}}")
 
-	s, err := mustache.Render(s, true, ctx)
+	s, err := mustache.RenderRaw(s, true, ctx)
```
shawn-hurley, line 614: Why render raw vs render just for my knowledge :)

shawn-hurley (approved):
Looks good to me, thank you for adding this!

provider/provider.go
jmle: We were calling the `Render` method passing `true` as a second parameter, but the signature of Render doesn´t have a bool parameter. `RenderRaw` has, and I thought there was some reason in the past to use `RenderRaw` but the invocation was done mistakenly by calling `Render` instead...

shawn-hurley: Thanks! that makes sense

PR #723, opened 2024-11-15
:sparkles: Adding a Scope to the running of the rule engine

shawn-hurley:
The scope will allow for users to call the engine directly will be able to add information to the condition context. Providers are still responsible for restricting based on that condition context.

Working off of @jmle awesome work to get the file paths respected in providers

### DIFF
--- a/engine/scopes.go
+++ b/engine/scopes.go
@@ -1,0 +1,88 @@
+package engine
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+)
+
+const TemplateContextPathScopeKey = "konveyor.io/path-scope"
+
+// Scopes apply to individual calls to the providers and will add inforamtion to the ConditionContext
+// To apply the scope. It is the responsiblity of the provider to use these correctly.
+type Scope interface {
+	Name() string
+	// For now this is the only place that we are considering adding a scope
+	// in the future, we could scope other things
+	AddToContext(*ConditionContext) error
+}
+
+type scopeWrapper struct {
+	scopes []Scope
+}
+
+func (s *scopeWrapper) Name() string {
+	name := ""
+	for i, s := range s.scopes {
+		if i == 0 {
+			name = s.Name()
+
+		} else {
+			name = fmt.Sprintf("%s -- %s", name, s.Name())
+		}
+	}
+	return name
+}
+
+func (s *scopeWrapper) AddToContext(conditionCTX *ConditionContext) error {
+	for _, s := range s.scopes {
+		err := s.AddToContext(conditionCTX)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+var _ Scope = &scopeWrapper{}
+
+func NewScope(scopes ...Scope) Scope {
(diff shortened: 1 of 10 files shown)


provider/lib.go
```diff
@@ -71,9 +76,15 @@ func GetFiles(configLocation string, filepaths []string, patterns ...string) ([]
 		}
 	} else if len(filepaths) == 1 {
 		// Currently, rendering will render a list as a space separated paths as a single string.
+		log.Info("here file path 1")
 		patterns := strings.Split(filepaths[0], " ")
 		for _, pattern := range patterns {
+			if p, err := filepath.Rel(configLocation, pattern); err == nil {
```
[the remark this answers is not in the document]
shawn-hurley: Updated to use the full file path when asking if it matches the pattern.

engine/scopes.go
```diff
@@ -0,0 +1,88 @@
...
+
+var _ Scope = &includedPathScope{}
+
+func (i *includedPathScope) Name() string {
+	return "IncludedPathScope"
+}
+
+// This will only update conditionCTX if filepaths is not set.
+func (i *includedPathScope) AddToContext(conditionCTX *ConditionContext) error {
+	// If any chain template has the filepaths set, only use those.
```
jmle, line 66: I guess in this case we are fine with scoping with the existing filepaths, is that the reason for this?

provider/internal/builtin/service_client.go
```diff
@@ -47,16 +47,38 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
...
+					matched, err = filepath.Match(c.Pattern, path)
+					if err != nil {
+						continue
+					}
+				}
+				if matched {
+					matchingFiles = append(matchingFiles, path)
+				}
+			}
+		} else {
```
jmle, line 77: The problem here is that if we have a rule with filepaths and scopes are also added, its filepaths are ignored? That would essentially change the rule. Sorry if I'm not understanding this correctly.

jmle (approved):

jmle (changes requested):
Just some questions

engine/scopes.go
shawn-hurley: Yeah, say for instance that you scope to pom.xml, but the condition is looking at /resources/web. in this case, we want the engine to more or less ignore this rule; we are saying in the scope that it only cares about pom.xml and only gives me results related to that file.

Does that help?

provider/internal/builtin/service_client.go
shawn-hurley: I tried to answer this question above as well as that question, but you are basically on point. 

say a rule is searching for web.xml (hardcode or found) but when we set the scope of the analysis, we only want pom.xml. In this case I **think** the right thing to do, is only search the filepaths that is the intersection of scope of the analysis and the filepaths from the rule. 

Basically I imagine the scope, is as if you copied the included paths to $PLACE and only ran analysis on $PLACE. 

Does that help answer the question?

jmle: Makes sense, thanks!

PR #737, opened 2024-11-22
:sparkles: Add go grep package to replace OS-specific grep commands for builtin provider

kthatipally:
### PR Summary 
This pull request introduces a cross-platform supported grep package to replace OS-specific grep commands. 
This the regex2 package to scan files and search for patterns, making the codebase more portable and easier to maintain: https://github.com/dlclark/regexp2 (supports lookahead)

provider/internal/builtin/service_client.go: Replaced OS-specific grep command with a cross-platform solution using regexp2 and file scanning.
go.mod and go.sum: Added regexp2 dependency.

WIP: Testing, resolve merge-conflicts

### DIFF
--- a/provider/internal/builtin/service_client.go
+++ b/provider/internal/builtin/service_client.go
@@ -1,24 +1,26 @@
 package builtin
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io/fs"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"regexp"
-	"strconv"
 	"strings"
 	"sync"
 
 	"github.com/antchfx/jsonquery"
 	"github.com/antchfx/xmlquery"
 	"github.com/antchfx/xpath"
+	"github.com/dlclark/regexp2"
 	"github.com/go-logr/logr"
+	"github.com/konveyor/analyzer-lsp/lsp/protocol"
 	"github.com/konveyor/analyzer-lsp/provider"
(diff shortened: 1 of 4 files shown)


provider/internal/builtin/service_client.go
```diff
@@ -497,3 +473,91 @@ func (b *builtinServiceClient) isFileIncluded(absolutePath string) bool {
...
+	var positionsMu sync.Mutex
+	var eg errgroup.Group
+
+	err := filepath.Walk(location, func(path string, f os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if f.Mode().IsRegular() {
+			eg.Go(func() error {
```
shawn-hurley, line 493: We might want to put a limit on this across the engine this could spawn a bunch of go routines I think just setting the limit on the errorgroup.Group should be enough

shawn-hurley:
We need to test this with this rule id to make sure that it covers the lookahead and lookbehind 

https://github.com/konveyor/rulesets/blob/f7ce5496f11d73d40c126a8d9be5f3880481a2a3/default/generated/eap7/108-eap6.windup.yaml#L112

shawn-hurley:
@kthatipally I have updated the pr here #810 to handle the merge conflicts and add in a little change that will allow us to more quickly find files that don't match. I think this gets us back to where we were. let me know if it looks good to you or if it is ok to close this in favor of that.

PR #738, opened 2024-11-25
:bug: CP # 711- fix builtin provider filecontent cond for darwin

eemcmullan:


provider/internal/builtin/service_client.go
```diff
@@ -84,15 +85,25 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
 			return response, fmt.Errorf("could not parse provided regex pattern as string: %v", conditionInfo)
 		}
 		var outputBytes []byte
-		grep := exec.Command("grep", "-o", "-n", "-R", "-P", c.Pattern, p.config.Location)
-		outputBytes, err := grep.Output()
+		var err error
+		if runtime.GOOS == "darwin" {
+			cmd := fmt.Sprintf(
+				`find %v -type f | \
+			while read file; do perl -ne '/(%v)/ && print "$ARGV:$.:$1\n";' "$file"; done`,
```
jmle, line 92: I guess this is because of the differences in the installed grep, BSD vs GNU? Actually, relying on external tools to do the work when we're not running inside a container anymore might not be a good idea. Maybe we should think about doing this programmatically at some point.

jmle:
Looks good :+1:

aufi (approved):

PR #741, opened 2024-12-03
:bug: Fix perl command for MacOS

eemcmullan:


provider/internal/builtin/service_client.go
```diff
@@ -642,3 +653,12 @@ func runOSSpecificGrepCommand(pattern string, location string, providerContext p
 
 	return outputBytes, nil
 }
+
+func isSlashEscaped(str string) bool {
+	for i := 0; i < len(str); i++ {
+		if str[i] == '/' && i > 0 && str[i-1] == '\\' {
```
shawn-hurley, line 659: would this have to be str[i-2:i] to get both \\ because otherwise you would only get \ right?

eemcmullan: @shawn-hurley this only represents one escape char - two are needed here as one is an escape to check for that specific character in the string.

PR #742, opened 2024-12-03
:sparkles:  CP #717 - Allow configurable path for fernflower

eemcmullan:


shawn-hurley (approved):
I don't see anything wrong, 

The way we constantly have to change the arguments to the helper functions for things like this, starts to make me think that they should be a struct and that we should refactor but out of scope of this change.

dymurray (approved):

PR #746, opened 2024-12-10
:bug: Adding ability for provider and service client to shutdown gracefully

shawn-hurley:
This will update and catch the context cancel to close the pipes for stdin and stdout.

PR #748, opened 2024-12-11
:bug: CP # 746 - Adding ability for provider and service client to shutdown grac…

eemcmullan:
…efully (#746)

This will update and catch the context cancel to close the pipes for stdin and stdout.


shawn-hurley (approved):
+1 sorry for missing the cherry pick!

PR #759, opened 2025-01-10
:sparkles: Exclusion support [Part 2]

pranavgaikwad:
This is continuation of my previous PR for exclusion


eemcmullan (approved):

provider/internal/builtin/service_client.go
```diff
@@ -623,21 +651,50 @@ func runOSSpecificGrepCommand(pattern string, location string, providerContext p
...
 		escapedPattern = strings.ReplaceAll(escapedPattern, "$", "\\$")
-		cmd := fmt.Sprintf(
-			`find %v -type f -exec perl -ne 'print "$ARGV:$.:$1\n" if /%v/; close ARGV if eof;' {} +`,
-			location, escapedPattern,
-		)
+		cmd := ""
+		if ok, paths := providerContext.GetScopedFilepaths(); ok {
+			cmd = fmt.Sprintf(
+				`echo '%s' | \
+			xargs perl -ne '/%v/ && print "$ARGV:$.:$1\n";'`,
```
aufi, line 658: Just curious if `xargs perl` or `find` on following lines behave as we expect on other plaforms like Mac (or also Windows?).

aufi (approved):

provider/internal/builtin/service_client.go
pranavgaikwad: perl is only for Mac, notice the if condition. We now have tests that run on all three platforms expecting identical output for same tcs

shawn-hurley:
@pranavgaikwad I think that #790 can help speed up the windows bit a little.

PR #760, opened 2025-01-14
:bug: make sure non java files are exploded to java project

pranavgaikwad:
API Tests PR: 249


eemcmullan (approved):

aufi (approved):
:+1:

aufi:
Regarding to test failures, failed acmeair-webapp and tackle-testapp-binary added a issue in properties file which looks good and API tests should be updated (as well as Tags) in follow-up PR to go-konveyor-tests.

However, administracion-efectivo cannot find any issue now (local-storage-00001 and local-storage-00002 from `.java` files were expected), which looks strange and probabaly needs some care on analyzer side (test case: https://github.com/konveyor/go-konveyor-tests/blob/main/analysis/tc_administracion_efectivo_upload_binary.go)

PR #770, opened 2025-02-06
:bug: handle concurrency issue with updating ruleCTX with scopes

shawn-hurley:
* Adding context propagation so that rule workers when processing A rule can attach their span to the correct parent

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -14,7 +14,9 @@ import (
 	"sync/atomic"
 
 	"go.lsp.dev/uri"
+	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/propagation"
 
 	"github.com/cbroglie/mustache"
 	"github.com/go-logr/logr"
@@ -31,11 +33,12 @@ type RuleEngine interface {
 }
 
 type ruleMessage struct {
-	rule        Rule
-	ruleSetName string
-	ctx         ConditionContext
-	scope       Scope
-	returnChan  chan response
+	rule             Rule
+	ruleSetName      string
+	conditionContext ConditionContext
+	scope            Scope
+	returnChan       chan response
+	carrier          propagation.TextMapCarrier
 }
 
 type response struct {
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -190,11 +196,15 @@ func (r *ruleEngine) RunRulesScoped(ctx context.Context, ruleSets []RuleSet, sco
...
 	}
+	carrier := propagation.MapCarrier{}
+	otel.GetTextMapPropagator().Inject(ctx, carrier)
+	r.logger.Info("inject span info", "carrier", carrier)
 	ctx, cancelFunc := context.WithCancel(ctx)
 
 	taggingRules, otherRules, mapRuleSets := r.filterRules(ruleSets, selectors...)
 
 	ruleContext := r.runTaggingRules(ctx, taggingRules, mapRuleSets, conditionContext, scopes)
+	r.logger.Info("finished taging rules")
```
fabianvf, line 207:
```suggestion
	r.logger.Info("finished tagging rules")
```

engine/engine.go
```diff
@@ -264,10 +274,13 @@ func (r *ruleEngine) RunRulesScoped(ctx context.Context, ruleSets []RuleSet, sco
 	}()
 
 	for _, rule := range otherRules {
+		newContext := ruleContext.Copy()
```
shawn-hurley, line 277: This was needed because otherwise, scopes were added in parallel, and this caused race conditions.

engine/engine.go
```diff
@@ -264,10 +274,13 @@ func (r *ruleEngine) RunRulesScoped(ctx context.Context, ruleSets []RuleSet, sco
...
+		newContext.RuleID = rule.rule.RuleID
 		wg.Add(1)
 		rule.returnChan = ret
-		rule.ctx = ruleContext
+		rule.conditionContext = newContext
```
shawn-hurley, line 281: renamed for claritiy

fabianvf (approved):

pranavgaikwad (approved):

PR #777, opened 2025-02-18
:sparkles: Adding ability for the cache for dependencies to re-gen

shawn-hurley:
* When the pom file has changed (by comparing a hash of the file) we will regen the cache.
* this should work for all the cases that have a pom
* there is a TODO to add this behavior for gradle, but that didn't ever seem to be cached.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/dependency.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/dependency.go
@@ -112,35 +112,16 @@ func (p *javaServiceClient) GetDependencies(ctx context.Context) (map[uri.URI][]
 		return m, err
 	}
 
-	if pomFile := p.findPom(); pomFile != "" {
-		// Read pom and create a hash.
-		// if pom hash and depCache return cache
-		file, err := os.Open(pomFile)
-		if err != nil {
-			p.log.Error(err, "unable to open the pom file", "pom path", pomFile)
-			return nil, err
-		}
-		hash := sha256.New()
-		if _, err := io.Copy(hash, file); err != nil {
-			p.log.Error(err, "unable to copy file to hash", "pom path", pomFile)
-			return nil, err
-		}
-		hashString := string(hash.Sum(nil))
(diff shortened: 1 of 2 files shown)


eemcmullan (approved):

external-providers/java-external-provider/pkg/java_external_provider/dependency.go
```diff
@@ -109,11 +112,29 @@ func (p *javaServiceClient) GetDependencies(ctx context.Context) (map[uri.URI][]
...
+	if pomFile := p.findPom(); pomFile != "" {
+		// Read pom and create a hash.
+		// if pom hash and depCache return cache
+		file, err := os.Open(pomFile)
+		if err != nil {
+			p.log.Error(err, "unable to open the pom file", "pom path", pomFile)
+			return nil, err
+		}
+		hash := sha256.New()
+		if _, err := io.Copy(hash, file); err != nil {
```
eemcmullan: Looks like `file` may not always be a file? From demo output: `0.errors.java-pomxml-dependencies = read /analyzer-lsp/examples/quarkus-1-6-2-jar-exploded: is a directory`

shawn-hurley: Yeah, I think in the case of binaries, there is some special stuff, I don't need the behavior changed for that, so adding back the behavior for that and adding handling nested poms.

eemcmullan (approved):

fabianvf (approved):

shawn-hurley:
Something that I am doing is now making the tests time out. I couldn't tell you what that is, though. This should behave the same besides the file path walk, and if we were really already that tight on time, then something else is wrong. 

I am going to need to dig in more here.

shawn-hurley:
@eemcmullan @fabianvf can you re-look this is now passing CI and it works for kai. Want to get another pair of eyes because i changed it quite a bit from approval.

eemcmullan (approved):

PR #780, opened 2025-02-20
:bug: Fixing how windows paths are created to regex

shawn-hurley:


eemcmullan (approved):
Thanks for finding and fixing!

PR #786, opened 2025-03-05
:bug: And and Or conditions could pass with exluded data.

shawn-hurley:
If the incidents were filtered out in the next step, the resulting complete response matched would still be set incorrectly.

We had to filter out not only the engine layer but the ConditionEntry.

### DIFF
--- a/engine/conditions.go
+++ b/engine/conditions.go
@@ -171,10 +171,11 @@ func (a AndCondition) Evaluate(ctx context.Context, log logr.Logger, condCtx Con
 			// TODO: determine if this is the right thing, I am assume the full rule should fail here
 			return ConditionResponse{}, fmt.Errorf("unable to find context value: %v", c.From)
 		}
-		response, err := c.ProviderSpecificConfig.Evaluate(ctx, log, condCtx)
+		response, err := c.Evaluate(ctx, log, condCtx)
 		if err != nil {
 			return ConditionResponse{}, err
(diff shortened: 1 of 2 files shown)


engine/conditions.go
```diff
@@ -171,22 +171,19 @@ func (a AndCondition) Evaluate(ctx context.Context, log logr.Logger, condCtx Con
...
 		}
+		// Must filter out based on the filepaths.
 		if c.As != "" {
 			condCtx.Template[c.As] = ChainTemplate{
 				Filepaths: incidentsToFilepaths(response.Incidents),
 				Extras:    response.TemplateContext,
 			}
 		}
 
-		matched := response.Matched
```
pranavgaikwad, line 185: I think I am not following why we remove 'not' here?

pranavgaikwad (commented):
Looks good, great find! Just one question:

engine/conditions.go
shawn-hurley: Because I switched this to use https://github.com/konveyor/analyzer-lsp/pull/786/files#diff-bd3e00a860d81ad866acc697ae8784bd65ecd03dc56d5b4643b4768234bba2b8R174 which contains this logic.

pranavgaikwad: awesome, I understand now

pranavgaikwad (approved):

PR #790, opened 2025-03-10
:bug: Removing for loops for the windows file content search

shawn-hurley:
* Seeing runs of kai on windows go from 2m6s to 1m20s
* Still need a full solution but is better then it was before


shawn-hurley:
removing in favor or removing the specific grep. #810

PR #796, opened 2025-03-14
:sparkles: add rule progress logs

zhoufenqin:
As [this issue](https://github.com/konveyor/kantra/issues/444) in kantra, it's better to show to log processing in the console so user can know how many rules has been processed.
we can specify the log with "process-rule" name so that the kantra CLI can use log hook to select logs print in console.
related PR in kantra: https://github.com/konveyor/kantra/pull/453

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -459,6 +459,7 @@ func parseTagsFromPerformString(tagString string) ([]string, error) {
 }
 
 func processRule(ctx context.Context, rule Rule, ruleCtx ConditionContext, log logr.Logger) (ConditionResponse, error) {
+	log.WithName("process-rule").Info("processing rule", "ruleID", rule.RuleID)
 	ctx, span := tracing.StartNewSpan(
 		ctx, "process-rule", attribute.Key("rule").String(rule.RuleID))
 	defer span.End()


zhoufenqin:
@shawn-hurley Thank you for your explain, I didn't choose to update the log to stdout instand of anslysis.log file becuase the log is too verbose. 

```
The engine can't put things directly onto stdout because that would cause problems for other use-cases (kai_analyzer_rpc specifically)
```
For this reason, I don't have more context, can you explain more about it or is there any record issue before?
Besides, I don't use the engion to directly put the log onto stdout, I still put the log to the path configured by the logger, then use log hook to print the log into the stdout in [kantra](https://github.com/konveyor/kantra/pull/453) repo, do you mean this kind of opeation may cause problems for other use-cases (kai_analyzer_rpc specifically)

```
we could have an engine option for that accepts a "stdout" logger
```

Yes, if you can, it's better, let me understand the step:
1. accept the stdout logger
``` go
type ruleEngine struct {
	// Buffered channel where Rule Processors are watching
	ruleProcessing chan ruleMessage
	cancelFunc     context.CancelFunc
	logger         logr.Logger
        **consoleLogger  logr.Logger**  // Add a logger here

	wg *sync.WaitGroup

	incidentLimit    int
	codeSnipLimit    int
	contextLines     int
	incidentSelector string
	locationPrefixes []string
}
```
2.  log the rule processing in the same location.

Add the rule processing log in `processRule` function, then there is no need to use log hook in kantra
```
consoleLogger.WithName("process-rule").Info("processing rule", "ruleID", rule.RuleID) 
```

shawn-hurley:
@zhoufenqin I am sorry, I have no idea what I was looking at that was making me concerned because you right this is would just hook into the logger. 

I think this change makes sense sorry about that.

shawn-hurley (approved):

zhoufenqin:
thanks @shawn-hurley, I also update in kantra repo, we discussed under that pr, see https://github.com/konveyor/kantra/pull/453, if it's reasonable, can you help to approve and merge?  I don't have permission to merge pr  :)

PR #802, opened 2025-03-24
:sparkles: Removed Maven Search and build up dep info from jar

shawn-hurley:
* Removed searching for dep by SHA from maven search
* Attempt to build up the dependency information from well known MANIFEST.MF headers, handle OSGI and base jar headers.
* Added a fallback to handle cases when we are unable to compute, this happens in the case of old JAR files that don't specifiy these values.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/service_client.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/service_client.go
@@ -225,9 +225,9 @@ func (p *javaServiceClient) Stop() {
 		p.log.Info("stopping java provider", "error", err)
 	}
 	if len(p.cleanExplodedBins) > 0 {
-		for _, explodedPath := range p.cleanExplodedBins {
-			os.RemoveAll(explodedPath)
-		}
+		//for _, explodedPath := range p.cleanExplodedBins {
+		//os.RemoveAll(explodedPath)
+		//}
 	}
 }
 
--- a/external-providers/java-external-provider/pkg/java_external_provider/util.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/util.go
@@ -4,13 +4,9 @@ import (
 	"archive/zip"
 	"bufio"
 	"context"
-	"crypto/sha1"
-	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"io"
 	"math"
-	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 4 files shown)


external-providers/java-external-provider/pkg/java_external_provider/service_client.go
```diff
@@ -225,9 +225,9 @@ func (p *javaServiceClient) Stop() {
 		p.log.Info("stopping java provider", "error", err)
 	}
 	if len(p.cleanExplodedBins) > 0 {
-		for _, explodedPath := range p.cleanExplodedBins {
-			os.RemoveAll(explodedPath)
-		}
+		//for _, explodedPath := range p.cleanExplodedBins {
```
eemcmullan, line 228: Do we want to keep the exploded paths after analysis?

external-providers/java-external-provider/pkg/java_external_provider/util.go
```diff
@@ -292,119 +292,111 @@ func explode(ctx context.Context, log logr.Logger, archivePath, projectPath stri
...
+				groupPath := filepath.Join(strings.Split(dep.GroupId, ".")...)
+				artifactPath := filepath.Join(dep.ArtifactId)
+				destPath := filepath.Join(m2Repo, groupPath, artifactPath,
+					dep.Version, fmt.Sprintf("%s-%s.jar", dep.ArtifactId, dep.Version))
+				if err := CopyFile(explodedFilePath, destPath); err != nil {
+					log.V(8).Error(err, "failed copying jar to m2 local repo")
+				} else {
+					log.V(8).Info("copied jar file", "src", explodedFilePath, "dest", destPath)
+				}
+			} else {
```
eemcmullan, line 377: Is this to show that we found a dep, but not all of the info for it?

external-providers/java-external-provider/pkg/java_external_provider/util.go
```diff
@@ -525,102 +521,131 @@ func constructArtifactFromPom(jarFile string) (javaArtifact, error) {
...
 	}
+	defer jar.Close()
 
-	var data map[string]interface{}
-	err = json.Unmarshal(body, &data)
-	if err != nil {
-		return dep, err
-	}
+	for _, file := range jar.File {
+		match, err := filepath.Match("META-INF/maven/*/*/pom.properties", file.Name)
```
eemcmullan, line 619: Would this path cause an issue in Windows?

shawn-hurley:
I am closing this PR, because I think that we need to create an enhancement and track this work a little bit better.

PR #807, opened 2025-04-02
:bug: nil check for ref.filepaths & downgrade some deps until go 1.23

eemcmullan:


aufi (approved):

external-providers/java-external-provider/pkg/java_external_provider/service_client.go
```diff
@@ -57,7 +57,7 @@ func (p *javaServiceClient) Evaluate(ctx context.Context, cap string, conditionI
 		return provider.ProviderEvaluateResponse{}, fmt.Errorf("unable to get query info: %v", err)
 	}
 	// filepaths get rendered as a string and must be converted
-	if len(cond.Referenced.Filepaths) > 0 {
+	if cond.Referenced.Filepaths != nil && len(cond.Referenced.Filepaths) > 0 {
```
pranavgaikwad, line 60: doesn't len() do an implicit nil check?

eemcmullan: I thought so - but I was hitting a nil panic error on this line in kantra. After adding this in, it fixed that.

pranavgaikwad (approved):

PR #811, opened 2025-04-07
:sparkles: parallelize tagging rules

pranavgaikwad:
API Tests PR: 269

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -179,12 +179,11 @@ func (r *ruleEngine) RunRules(ctx context.Context, ruleSets []RuleSet, selectors
 }
 
 func (r *ruleEngine) RunRulesScoped(ctx context.Context, ruleSets []RuleSet, scopes Scope, selectors ...RuleSelector) []konveyor.RuleSet {
-	// determine if we should run
-
 	conditionContext := ConditionContext{
 		Tags:     make(map[string]interface{}),
 		Template: make(map[string]ChainTemplate),
 	}
+
 	if scopes != nil {
 		r.logger.Info("using scopes", "scope", scopes.Name())
 		err := scopes.AddToContext(&conditionContext)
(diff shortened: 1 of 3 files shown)


pranavgaikwad:
This is fixing previously unknown bugs in that we are now getting more tags because the dependent tagging rules (ones with hasTags) are now processed in correct order i.e. after normal tagging rules. I am updating the tests.

engine/engine.go
```diff
@@ -196,125 +195,256 @@ func (r *ruleEngine) RunRulesScoped(ctx context.Context, ruleSets []RuleSet, sco
...
+							return
+						}
+						atomic.AddInt32(&matchedRules, 1)
+						rs, ok := rulesetsMap[response.RuleSetName]
+						if !ok {
+							r.logger.Info("this should never happen that we don't find the ruleset")
+							return
+						}
+						if response.Rule.Perform.Tag != nil {
+							// handle a tagging rule (info rule)
```
jmle, line 338: Just a suggestion: I would move this to a different method, since it seems to me a more specific concern

engine/engine.go
```diff
@@ -196,125 +195,256 @@ func (r *ruleEngine) RunRulesScoped(ctx context.Context, ruleSets []RuleSet, sco
...
+		int32(totalRules),
+		violationRules,
+		scopes,
+		conditionContext,
+		carrier,
+		rulesetsMap,
+	)
+	if err != nil {
+		r.logger.Error(err, "context canceled when processing violation rules")
+	}
```
jmle, line 258: Suggestion: probably the three blocks above can be extracted to a method

jmle (approved):
Looks good, I just left some readability suggestions. I think this could fix quite a few problems with consistency :+1: :+1:

engine/engine.go
```diff
@@ -196,125 +195,256 @@ func (r *ruleEngine) RunRulesScoped(ctx context.Context, ruleSets []RuleSet, sco
...
+	}()
+
 	// Handle returns
 	go func() {
 		for {
 			select {
-			case response := <-ret:
+			case response := <-ruleResponseChan:
 				func() {
 					r.logger.Info("rule returned", "ruleID", response.Rule.RuleID)
```
shawn-hurley, line 308: Starting feel like this should be it's own method not inline

shawn-hurley (approved):

PR #814, opened 2025-04-11
:sparkles: Feature/document change listeners

pranavgaikwad:
API Tests PR: 270

This PR introduces new API NotifyFileChanges() in the providers. It is supposed to be used to send changes to files that happen in the editor. 

Currently, the api is implemented by the builtin provider where - it introduces a new working copy manager that listens for file changes and writes changes to a temp location. It then uses these new paths to make additional search and excludes the original paths. As soon as a file is saved, the temp copies are discarded. 

It uses inclusion / exclusion constraints dynamically; I also refactored how we were using include / exclude and doing file searches as the logic was duplicated in different conditions. I consolidated all that under FileSearcher


fabianvf (approved):
This is clean, I like it. I saw a few TODOs/unimplemented change listeners, if that's not a big deal then LGTM

PR #815, opened 2025-04-13
:sparkles: add ability for java service client to use rpc passed to it for

shawn-hurley:


external-providers/java-external-provider/pkg/java_external_provider/provider.go
```diff
@@ -228,6 +228,16 @@ func (p *javaProvider) Init(ctx context.Context, log logr.Logger, config provide
 	}
 	log = log.WithValues("provider", "java")
 
+	if config.RPC != nil {
+		return &javaServiceClient{
+			rpc:               config.RPC,
```
jmle, line 233: What will this config look like? A sample somewhere could be useful

jmle (approved):
Looks good :+1:

external-providers/java-external-provider/pkg/java_external_provider/provider.go
shawn-hurley: It is only for someone using this as library and calling java-external-provider from go code. I expect the use case to be limited to us (kai/kantra) and I don't expect it to be a user facing change. 

Thoughts on where/how this should be documented in that case?

jmle: If that's the case then it's cool :+1:

pranavgaikwad (approved):

PR #823, opened 2025-05-09
:bug: NodeJS provider time for conn.Notify to process

eemcmullan:


aufi (approved):
2 seconds per-file looks quite long time, however LGTM

eemcmullan:
@aufi I agree. I think for 0.8, the Node service client will need to be refactored for better performance.

PR #826, opened 2025-05-14
:bug: MTA-5156 block GetDependency() calls in java until previous one is finished

pranavgaikwad:


jmle (changes requested):
Looks good, I am guessing that with nexus being a multi-module project we were somehow triggering several `mvn dependency:tree` invocations? We could actually do a single invocation to get all dependencies, with `mvn dependency:tree -pl module1,module2`. We would have to get the modules previously, but it could also speed things up. Might be worth leaving a TODO.

In any case it looks like there is a problem with one dependency missing in a test.

jmle (approved):
LGTM

eemcmullan (approved):

PR #832, opened 2025-05-16
:bug: Fix for .Net provider URI for Windows

eemcmullan:


aufi (approved):

external-providers/dotnet-external-provider/pkg/dotnet/service_client.go
```diff
@@ -168,6 +169,11 @@ func processFile(path string, regex *regexp.Regexp, positionsChan chan<- interfa
 				if err != nil {
 					return
 				}
+
+				if runtime.GOOS == "windows" {
+					absPath = strings.ReplaceAll(absPath, "\\", "/")
```
djzager, line 174: Why not this?

```suggestion
					absPath = filepath.ToSlash(absPath)
```

djzager (commented):
Just a thought. I think this is better choice but I don't want to block without having more time to research.

PR #841, opened 2025-06-10
:bug: MTA-5369 Revert ":sparkles: parallelize tagging rules"

pranavgaikwad:
Reverts konveyor/analyzer-lsp#811


zhoufenqin:
Hi, why revert the change?

PR #844, opened 2025-06-12
:ghost: fix workflows

pranavgaikwad:


.github/workflows/demo-testing.yml
```diff
@@ -1,12 +1,18 @@
 name: Demo Testing
 
-on: ["push", "pull_request"]
+on:
```
pranavgaikwad, line 3: this was needed because we were running push on all branches including revert and cherry-picks created by automation

aufi (approved):
Looks great.

jmle (approved):

PR #867, opened 2025-08-01
Attempting to break apart and encapsulate logic in java provider/svcclient and dependency

shawn-hurley:
* We have many different types of run modes and behaviour is subtly
  dependent on each other.
* In this approach, everything will be tied to gether that needs to be
  and all the subtly will be explict, such as resolving sources must
come after decompiling a binary.
* When decompiling we have to take into account each type of archive and
  where they should be putting files
* When resolving sources, we need specific build tool version.
* Right now, during init, we do all of this work with deps, and then
  IIUC re-do it in the dependency code, We shouldn't need to do this I
think.

* I also want to change the decompile behaviour because fernflower can
  decompile the whole archive, we don't have to go file by file and
decompile each class.

Signed-off-by: Shawn Hurley <shawn@hurley.page>

## Summary by CodeRabbit

* **New Features**
  * Added detailed documentation outlining the initialization flow for the Java external provider.
  * Introduced functionality to start and manage the Eclipse JDT Language Server for Java development.
  * Implemented new abstractions and resolvers for handling and decompiling Java sources, including Maven and Gradle support.
  * Enhanced dependency identification and labeling during Java archive decompilation, including improved fallback mechanisms.

* **Refactor**
  * Simplified and modularized Java provider initialization for better maintainability and concurrency.
  * Improved handling of Maven artifact downloads and proxy configuration.
  * Streamlined service client lifecycle management and dependency label integration.

* **Bug Fixes**
  * Improved fallback and error handling for incomplete or missing Java dependency metadata.

* **Documentation**
  * Added a comprehensive markdown file describing the internal initialization logic and known issues.

* **Chores**
  * Reordered sample rule entries in the demo output for improved organization.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/dependency.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/dependency.go
@@ -2,7 +2,6 @@ package java
 
 import (
 	"bufio"
-	"bytes"
 	"context"
 	"crypto/sha256"
 	"errors"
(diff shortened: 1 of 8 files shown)


shawn-hurley:
This is captured and superceded by #992 closing

PR #873, opened 2025-08-07
:bug: Fixing how we determine deps and what we do with it

shawn-hurley:
* refactoring the code a little bit.
* adding back the idea of "foundOnline" which we may want to consider renaming.
* adding "is valid" because we really only care if we can create mvn coordinates.
* Adding logs at the base level to make debugging easier.
* Removed a case where we tried to handle if dep was valid but got an error. I don't see how that is possible and I worried that it created confusion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of Java dependencies to ensure only valid artifacts are processed.
  * Enhanced error handling and logging for dependency resolution and decompilation steps.

* **Refactor**
  * Streamlined the process for handling nested JAR files and dependency resolution, resulting in clearer logging and more robust fallback mechanisms.

* **Style**
  * Adjusted logging levels for better clarity and reduced verbosity in certain scenarios.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/util.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/util.go
@@ -24,6 +24,8 @@ import (
 	"math/rand"
 
 	"github.com/go-logr/logr"
+	"github.com/konveyor/analyzer-lsp/engine/labels"
+	"github.com/konveyor/analyzer-lsp/provider"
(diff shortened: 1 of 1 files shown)


shawn-hurley:
We will need to update test cases, these changes things in a real way, and can get better answers for some test cases.

eemcmullan (approved):

jmle (approved):
+100 on the refactor

Issue #795, opened 2025-03-13
[BUG] Xpath condition stops after finding a single match in a file


jmle:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

latest

### Priority

Critical

### Current Behavior

It looks like the builtin xpath condition stops looking into a file after it has found the first match, so if there are more potential matches in the file, they won't appear if a match was already found.

### Expected Behavior

The condition should match against all occurrences.

### How Reproducible

Always (Default)

### Environment

```markdown
- **Analyzer Image SHA** 
- **JDTLS  base image SHA**
- **Provider Settings JSON**
```

PR #887, opened 2025-08-19
:bug: Use XML line numbers directly from xmlquery

aufi:
There is a problem with line number identification for XML matched nodes (empty node, just with attributes and could be spanning multiple lines in the source XML file). This lead to find all relevant matches using xpath query, however all had the same incident line number, so we deleted as duplicate during the analysis.

The fix of the root cause of the issue is to set correct location of the matched element. It was complicated/close to impossible IMO to do it in reliable way with just update of current methods in analyzer.

In order to have XML Nodes correctly annotated with their line numbers in the source file, the `xmlquery` package was forked and updated. We should confirm if this is an acceptable way and potentialy try get this change into official upstream `antchf/xmlquery` repository, changes: https://github.com/antchfx/xmlquery/compare/master...aufi:xmlquery:line-numbers?expand=1

Fixes https://github.com/konveyor/analyzer-lsp/issues/795


aufi:
Waiting for feedback on upstream `xmlquery` PR https://github.com/antchfx/xmlquery/pull/132

aufi:
Need to sort out what demo test results changes and correct and what are failures.

PR #890, opened 2025-08-21
:bug: Do not add insight for tagging rule with effort

### LINKED ISSUE #889: Insight created for message & tagging rule with effort > 0

eemcmullan:
## Summary by CodeRabbit

* **New Features**
  * Findings with non-zero effort are now routed to Violations; zero/unspecified effort findings go to Insights. Tags are now attached both to rule metadata and to final findings for improved filtering and reporting.

* **Bug Fixes**
  * Preserves reported effort on findings instead of clearing it, ensuring accurate severity and classification.

Fixes #889

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -435,13 +435,23 @@ func (r *ruleEngine) runTaggingRules(ctx context.Context, infoRules []ruleMessag
 				mapRuleSets[ruleMessage.ruleSetName] = rs
 			}
 			if rs, ok := mapRuleSets[ruleMessage.ruleSetName]; ok {
-				violation.Effort = nil
 				violation.Category = nil
-				// we need to tie these incidents back to tags that created them
-				for tag := range tags {
-					violation.Labels = append(violation.Labels, fmt.Sprintf("tag=%s", tag))
+				if violation.Effort != nil && *violation.Effort > 0 {
+					// we need to tie these incidents back to tags that created them
+					for tag := range tags {
+						violation.Labels = append(violation.Labels, fmt.Sprintf("tag=%s", tag))
+					}
+					// don't create insight for effort > 0
+					rs.Violations[rule.RuleID] = violation
+				} else {
+					violation.Effort = nil
+					// we need to tie these incidents back to tags that created them
+					for tag := range tags {
+						violation.Labels = append(violation.Labels, fmt.Sprintf("tag=%s", tag))
+					}
+					rs.Insights[rule.RuleID] = violation
 				}
-				rs.Insights[rule.RuleID] = violation
+
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -435,13 +435,23 @@ func (r *ruleEngine) runTaggingRules(ctx context.Context, infoRules []ruleMessag
...
-				violation.Effort = nil
 				violation.Category = nil
-				// we need to tie these incidents back to tags that created them
-				for tag := range tags {
-					violation.Labels = append(violation.Labels, fmt.Sprintf("tag=%s", tag))
+				if violation.Effort != nil && *violation.Effort > 0 {
+					// we need to tie these incidents back to tags that created them
+					for tag := range tags {
+						violation.Labels = append(violation.Labels, fmt.Sprintf("tag=%s", tag))
+					}
```
jmle, line 443: Since this loop is repeated in both `if` and `else`, can it be taken out?

jmle (approved):
Looks good, just a nit

engine/engine.go
shawn-hurley: I think that the only difference AFAICT is which list on the ruleset you are putting them on, is that correct?

Also, I assume that these rules will be run again, because I think we put them in both tagging rules and non tagging rules.

I think that we should:
1. Make sure that if a rule has a tag it only is added to the tagging rules.
2. Keep the for loop at the top level
3. make an if statement for the violation effort, and add it to violation's then use `continue` to run the item in the for loop
4. set the insight

I think that will make the code a little easier to read. If you want a example let me know!

shawn-hurley (changes requested):

shawn-hurley (commented):
So the more I think about this, this is something that I am worried about. 

It appears that if the rule has both tags and effort, we are moving that from a tagging rule to a violation rule. The one issue with this is if another rule has a condition on if the tag existis, then there is now a timing issue between those two violation rules. 

We choose to run all the info/tagging rules syncrounsly to start to handle this timing issue. 

Additionally, I believe these tags will be missing from the ruleset because I don't think there is a mechanism in the rule return to handle adding tags to the ruleset for violation rules.

I think that we should leave this:

https://github.com/konveyor/analyzer-lsp/pull/890/files#diff-2e6c4934f63ff9b712c2c346b33036af4724adf70b0801fff9b74f71b37fcd89L344-L351

and then this should handle adding the violation's https://github.com/konveyor/analyzer-lsp/pull/890/files#diff-2e6c4934f63ff9b712c2c346b33036af4724adf70b0801fff9b74f71b37fcd89R444-R450

shawn-hurley (approved):
Everything looks good to me, just as follow on can we verify that we are not unintentionally breaking a UI feature?

Issue #884, opened 2025-08-19
[BUG] Dependency custom rule is not triggered if credencial is not attached


mguetta1:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

0.8.0-alpha.4

### Priority

Major

### Current Behavior

Dependency custom rule is not triggered if credencial is not attached to the application

### Expected Behavior

The custom rule should be triggered regardless existence of credencials

### How Reproducible

Always (Default)

### Steps To Reproduce

Analyze [tackle-test-public](https://github.com/konveyor/tackle-testapp-public) app with the following analysis configuration:
1. Source code + dependencies
2. Skip "Set targets"
3. Upload [this](https://gist.github.com/mguetta1/2805e72a09c7bd090e9a9ba91c80e5f9) as manual rules file
4. No additional advanced options

### Environment

```markdown

```

PR #891, opened 2025-08-25
:bug: Fallback dependency mechanism should not require version

jmle:
Dependencies without version in the pom.xml should also be included

Fixes https://github.com/konveyor/analyzer-lsp/issues/884 / https://issues.redhat.com/browse/MTA-5979

## Summary by CodeRabbit

* **Bug Fixes**
  * Fallback dependency collection now includes dependencies even when version information is missing, improving completeness of results.
  * Non-version fields (such as name and additional metadata) are still populated; version and related file link may be absent for those entries.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/dependency.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/dependency.go
@@ -309,7 +309,7 @@ func (p *javaServiceClient) GetDependenciesFallback(ctx context.Context, locatio
 
 	// add each dependency found
 	for _, d := range pomDeps {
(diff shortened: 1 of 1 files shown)


shawn-hurley:
@jmle, when does this occur? Shouldn't the effective pom file have the version information?

shawn-hurley:
So we are saying that: [dependnecy:tree](https://github.com/konveyor/analyzer-lsp/blob/main/external-providers/java-external-provider/pkg/java_external_provider/dependency.go#L402) doesn't follow the BOM file IIUC. I think that this is a larger problem if that is the case, and that we should ensure we get the version specified in the BOM file.

what are your thoughts?

aufi (approved):

PR #893, opened 2025-08-26
:bug: Cache Maven Search Err & Fix artifactID

eemcmullan:
## Summary by CodeRabbit

- Bug Fixes
  - Corrected artifact path resolution for exploded JARs, ensuring proper local repository layout and copy destinations.
- Reliability
  - More robust handling of Maven search outages and non-OK responses with clearer errors, reducing noisy failures and stabilizing dependency retrieval.
- Performance
  - Caches server-side Maven search errors to avoid repeated lookups, reducing unnecessary network calls and speeding up repeated operations.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/util.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/util.go
@@ -407,7 +407,7 @@ func explode(ctx context.Context, log logr.Logger, archivePath, projectPath stri
 				dependencies = append(dependencies, dep)
 				// copy this into m2 repo to avoid downloading again
 				groupPath := filepath.Join(strings.Split(dep.GroupId, ".")...)
-				artifactPath := filepath.Join(strings.Split(dep.ArtifactId, ".")...)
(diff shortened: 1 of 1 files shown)


aufi (approved):
I can understand this looks little risky, but makes sense IMO. When there is a 5xx HTTP failure on given analysis run (which could be also 504 gateway timeout), maven central will be set to be skipped.

external-providers/java-external-provider/pkg/java_external_provider/util.go
```diff
@@ -553,6 +553,8 @@ func toDependency(_ context.Context, log logr.Logger, depToLabels map[string]*de
 	return dep, err
 }
 
+var mavenSearchErrorCache error
```
shawn-hurley, line 556: Add this as something to take care of in a rewrite. I think this will be fine for now but I kind of agree with Code Rabbit and that we might want to do something different in the future

shawn-hurley (approved):

jmle (approved):

PR #921, opened 2025-10-16
:sparkles: Add Maven index search functionality

jmle:
- Adds search functionality using pre-generated, bundled index
- Removes unneeded functionality:
  - Maven central HTTP sha search
  - Search by list of public groupIds
  - `disableMavenSearch` is no longer used - to be removed in a subsequent PR?
  
Addresses https://issues.redhat.com/browse/MTA-6231
Requires https://github.com/konveyor/java-analyzer-bundle/pull/158
Bundle PR: 158

## Summary by CodeRabbit

* **New Features**
  * Added local Maven index lookup and utilities (index entry type and random-name helper) for resolving Java artifacts.

* **Improvements**
  * Dependency extraction, decompilation, and source resolution now use a configurable Maven index path for more reliable, faster resolution.

* **Removed**
  * Removed reliance on open-source dependency label lists and remote Maven HTTP lookup in resolution flows.

* **Refactor**
  * Public interfaces updated to accept and propagate a Maven index path instead of label-based inputs.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/provider.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/provider.go
@@ -286,7 +286,7 @@ func (p *javaProvider) Init(ctx context.Context, log logr.Logger, config provide
 	}
 
 	disableMavenSearch, ok := config.ProviderSpecificConfig[DISABLE_MAVEN_SEARCH].(bool)
-	mavenIndexPath, ok := config.ProviderSpecificConfig[providerSpecificConfigOpenSourceDepListKey].(string)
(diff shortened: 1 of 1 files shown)


shawn-hurley (changes requested):
After looking at the output of the files, couldn't you just do the binary search over the .txt file? it appears as if the file is sorted by the sha, and if that is the case then we can drop the 791MB .idx file.

shawn-hurley (approved):
This does conform to the enhancement which I did just agree to.

We should have follow-ons for this work, so going to approve this

PR #944, opened 2025-10-28
:bug: Fix ARG_MAX errors in builtin provider and add default excludes

tsanders-rh:
## Problem

When analyzing JavaScript/TypeScript projects with `node_modules` installed, the builtin provider fails with "argument list too long" errors. This occurs because the grep command receives all matching files (30,000+ in a typical node_modules) as command-line arguments, exceeding the OS ARG_MAX limit.

**Example error:**
```
failed to perform file content search - could not run grep with provided pattern
fork/exec /usr/bin/grep: argument list too long
```

This affects any project with large dependency directories, making the analyzer unusable for common JavaScript/TypeScript workflows.

## Root Cause

In `provider/internal/builtin/service_client.go`, the Linux grep implementation passes all file paths directly as command arguments:

```go
args = append(args, locations...)  // Can be 30,000+ files
cmd := exec.Command("grep", args...)
```

OS systems limit total argument length (typically 1-2 MB). With 30,000 files × ~80 chars per path = ~2.4 MB, this exceeds ARG_MAX.

### 1. Use xargs for file list (fixes ARG_MAX)

Changed Linux grep to match the macOS approach - pipe file list via stdin using xargs instead of command arguments:

```go
// Build null-separated file list
var fileList bytes.Buffer
for _, f := range currBatch {
    fileList.WriteString(f)
    fileList.WriteByte('\x00')
}

// Use xargs to avoid ARG_MAX limits
cmd := exec.Command("/bin/sh", "-c", "xargs -0 grep -o -n --with-filename -P 'pattern'")
cmd.Stdin = &fileList
```

This eliminates ARG_MAX issues entirely, as xargs automatically batches files.

### 2. Add sensible default excludes (prevents issue)

Added default excluded directories in `provider/lib.go`:
- `node_modules` (JavaScript/TypeScript)
- `vendor` (PHP/Go)
- `.git`
- `dist`, `build`, `target` (build outputs)
- `venv`, `.venv` (Python)

These directories are now excluded by default, preventing most users from hitting the ARG_MAX issue. Users can still add custom excludes via `providerSpecificConfig.excludedDirs`.

### 3. Document excludedDirs configuration

Updated `docs/providers.md` to document:
- The `excludedDirs` configuration option
- Default excluded directories
- How to add custom excludes

## Testing

**Test environment:**
- Project with 2 source files
- 34,423 files in node_modules
- ARG_MAX limit: 1,048,576 bytes

**Results:**
- ✅ Before fix: "argument list too long" error
- ✅ After fix: Analysis completes successfully, node_modules excluded by default
- ✅ All existing unit tests pass (27 tests in builtin provider)

## Impact

- Fixes immediate blocker for JS/TS project analysis
- Prevents 95% of cases via default excludes
- Maintains backward compatibility (user excludes still work)
- Improves performance by skipping dependency directories

## Related

This issue was discovered while testing PatternFly v5→v6 migration rules on JavaScript/TypeScript projects. The xargs approach matches the existing macOS implementation pattern in the codebase.

## Summary by CodeRabbit

* **New Features**
  * Added excludedDirs config to customize directory exclusions during analysis, with clear semantics for empty, non-empty, and absent settings.

* **Documentation**
  * Docs updated with examples for absolute/relative paths, pattern matching, default excludes, and guidance about performance when excluding dependency dirs.

* **Improvements**
  * Faster, more robust file-content scanning with a fast-path for small searches and improved batching/error handling for large projects.

* **Bug Fixes**
  * One test entry reclassified from error to unmatched.

### DIFF
--- a/provider/internal/builtin/service_client.go
+++ b/provider/internal/builtin/service_client.go
@@ -576,6 +576,7 @@ func (b *builtinServiceClient) performFileContentSearch(pattern string, location
 				}
 				// Escape pattern for safe shell interpolation
 				escapedPattern := strings.ReplaceAll(pattern, "'", "'\"'\"'")
+				// Use -- to mark end of options, preventing patterns like --pf- from being interpreted as options
 				cmdStr := fmt.Sprintf(
 					`xargs -0 grep -o -n --with-filename -P -- '%s'`,
 					escapedPattern,
@@ -588,14 +589,20 @@ func (b *builtinServiceClient) performFileContentSearch(pattern string, location
 			if err != nil {
 				if exitError, ok := err.(*exec.ExitError); ok {
 					// Exit code 1: grep found no matches
-					// Exit code 123: GNU xargs (Linux) exits with 123 when any invocation exits with 1-125
-					// When grep processes files across multiple xargs batches and some batches have matches
-					// while others don't, xargs will exit with 123 (not 1). The current code treats this as
-					// an error and discards the partial results in currOutput, causing false negatives.
-					// Apply this fix to handle both exit codes correctly:
-					if exitError.ExitCode() == 1 || exitError.ExitCode() == 123 {
-						err = nil // Clear error; treat as "no matches in this batch"
-						// Continue to next batch (don't return!)
+					if exitError.ExitCode() == 1 {
+						err = nil // Clear error; no matches in this batch
+					}
+					// Exit code 123: GNU xargs exits with 123 when any child exits with 1-125
+					// This includes both "no matches" (grep exit 1) and real errors (grep exit 2)
+					// Only clear 123 if stderr is empty; otherwise surface the error
+					if exitError.ExitCode() == 123 {
+						stderrStr := strings.TrimSpace(string(exitError.Stderr))
+						if stderrStr == "" {
+							err = nil // mixed batches, but no actual error output
+						} else {
+							// Real grep error (e.g., exit 2) - include stderr in error message
+							return nil, fmt.Errorf("could not run grep with provided pattern %+v; stderr=%s", err, stderrStr)
+						}
 					}
 				}
 				if err != nil {


shawn-hurley:
If you were to exclude all the node modules, then IIUC it would force `source-only` analysis. I generally agree that this should be the default, but we should have a way of searching for `full-analysis`. 

Can we piggyback on the FileSearcher we have and implement this in pure Go, removing the use of a platform-dependent tool?

tsanders-rh:
@shawn-hurley - We could instruct end users to run analysis just on /src, but will they just point at the whole repo and give up when it errors out.  I think sensible defaults is the way to go, with an ability to add custom or even remove all.   To support "analyze everything", the code could be modified to:

  Option 1: Empty array clears defaults
  {
    "providerSpecificConfig": {
      "excludedDirs": []  // Means "no excludes, not even defaults"
    }
  }

  Option 2: Explicit flag
  {
    "providerSpecificConfig": {
      "useDefaultExcludes": false,
      "excludedDirs": ["only-exclude-this"]
    }
  }

I can look into the FileSearcher if you think that is a better path forward.

shawn-hurley:
I really like option 1. 

IIUC, this would mean that the excludeDirs values mean:
1. nil = the default of source only (skip node_modules)
2. [] (or empty list) = nothing is excluded and the performance hit will be had.
3. ["some_path_to_exclude"] = only exclude the paths in the list.

With that change, we should use xargs in this PR, create a follow-up issue, and see if there is anything else we could do to address the other problems we may eventually need to work on.

shawn-hurley (approved):

shawn-hurley:
^ given the potential change in the meaning of the excludedDirs

tsanders-rh:
Here is the follow-up issue: https://github.com/konveyor/analyzer-lsp/issues/945

jmle (approved):

PR #957, opened 2025-11-07
:bug: Gracefully skip rules requiring unavailable providers

### LINKED ISSUE #956: Parser should gracefully skip rules requiring unavailable providers instead of logging errors
When analyzing a codebase, the rule parser logs ERROR messages for every rule that requires a provider that isn't running. This creates significant noise in the logs and confuses users.

## Current Behavior

Analyzing a TypeScript codebase (which only starts `nodejs` and `builtin` providers) produces 235+ error messages like:

```
ERRO[0007] unable to parse all the rules for ruleset error="unable to find provider for: java
unable to find provider for: java
unable to find provider for: java
...
(repeated 235+ times)
```

These errors come from default rulesets (azure, camel, eap*, etc.) that contain Java-specific rules.

## Expected Behavior

Rules requiring unavailable providers should be:
1. **Silently skipped** (or logged at DEBUG/V(5) level)
2. **Not treated as errors** - this is expected behavior when analyzing single-language codebases
3. **Allow other rules to load successfully** - partial ruleset loading should succeed

## Why This Matters

**Principle of least surprise:** When analyzing a TypeScript project, seeing hundreds of Java errors is confusing and looks like something is broken.

**Correct error classification:**
- ✅ **Skip silently:** Rule requires provider X, provider X not available
- ❌ **Error loudly:** Malformed YAML, syntax errors, invalid rule structure

**Benefits all consumers:** Any tool using analyzer-lsp (kantra, Tackle UI, custom tools) would benefit from cleaner output.

## Proposed Fix

In `parser/rule_parser.go`, when `LoadRules()` encounters a missing provider:

```go
internRuleSet, internNeedProviders, err := ruleParser.LoadRules(rulePath)
if err != nil {
    if isMissingProviderError(err) {
        // Log at debug level, skip these rules
        log.V(5).Info("skipped rules requiring unavailable provider", 
            "file", rulePath, 
            "provider", extractProviderName(err))
        // Return successfully parsed rules, ignore unparseable ones
        return partialRuleSet, availableProviders, nil
    } else {
        // Actual parsing error
        return nil, nil, err
    }
}
```

## Example Use Case

```bash
# Analyzing TypeScript codebase
kantra analyze --input /path/to/react-app

# Current: 235+ Java errors in logs
# Expected: Clean output, only relevant rules loaded
```

## Environment
- analyzer-lsp version: `v0.8.1-alpha.2` (and likely all versions)
- Issue occurs in: `parser/rule_parser.go` around line that returns `fmt.Errorf("unable to find provider for: %v", langProvider)`

## Related

This issue was discovered while testing nodejs provider integration with kantra and the new progress bar feature.

tsanders-rh:
When analyzing a codebase with a subset of providers (e.g., analyzing TypeScript code with only nodejs and builtin providers), rules that require unavailable providers (e.g., Java provider) would generate hundreds of error log messages like:

  "unable to find provider for: java"

This polluted logs and made it difficult to identify real issues.

This commit introduces a MissingProviderError type that allows the rule parser to distinguish between "provider not available" (which is expected and should be handled gracefully) and actual parsing errors (which should be logged).

Changes:
- Add MissingProviderError custom error type
- Update getConditionForProvider to return MissingProviderError
- Update getConditions and LoadRule to catch MissingProviderError and skip conditions/rules gracefully with debug-level logging
- Update "rule no provider" test to expect success (rule skipped) instead of error

Before: 217 error messages when analyzing tackle2-ui with nodejs
After: 0 error messages - rules are silently skipped at debug level

Fixes: konveyor/analyzer-lsp#956

## Summary by CodeRabbit

* **Bug Fixes**
  * Rules referencing unavailable providers are now detected and skipped with non-fatal logging instead of causing errors, improving resilience so other rules continue processing.
* **Tests**
  * Test suite updated to expect graceful skipping of rules when providers are missing rather than failing, reflecting the new non-fatal behavior.

### DIFF
--- a/parser/rule_parser.go
+++ b/parser/rule_parser.go
@@ -23,6 +23,15 @@ var defaultRuleSet = &engine.RuleSet{
 	Name: "konveyor-analysis",
 }
 
+// MissingProviderError indicates a rule requires a provider that is not available
+type MissingProviderError struct {
+	Provider string
+}
+
+func (e MissingProviderError) Error() string {
+	return fmt.Sprintf("unable to find provider for: %v", e.Provider)
+}
+
 type parserErrors struct {
 	errs []error
 }
@@ -348,7 +357,10 @@ func (r *RuleParser) LoadRule(filepath string) ([]engine.Rule, map[string]provid
 				}
 				conditions, provs, err := r.getConditions(m)
 				if err != nil {
-					r.Log.V(8).Error(err, "failed parsing conditions in or clause", "ruleID", ruleID, "file", filepath)
+					// Skip logging for missing provider errors - they're already logged at debug level
+					if _, ok := err.(MissingProviderError); !ok {
+						r.Log.V(8).Error(err, "failed parsing conditions in or clause", "ruleID", ruleID, "file", filepath)
+					}
 					return nil, nil, err
 				}
 				if len(conditions) == 0 {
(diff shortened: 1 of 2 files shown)


shawn-hurley (approved):
As we add more languages, this is going to be a real hassle because IMO we don't have a good ways of exposing rules/rulesets to the users. We have a lot of stuff that defaults to using the java rules and it's going to get confusing unless we do something.

PR #972, opened 2025-11-18
:bug: No need anymore to join paths

jmle:
API tests PR: 365

## Summary by CodeRabbit

* **Refactor**
  * Resolver now accepts an explicit Maven index file path for SHA-based artifact resolution and logging of local repo path was added.
* **Tests**
  * Tests updated to reference an explicit Maven index file for consistent index lookup.
* **Documentation**
  * Clarified that mavenIndexPath should point to the index file (maven-index.txt), not a directory.
* **Bug Fixes**
  * Return errors for unreadable index paths and resolve directory inputs to the index file.


shawn-hurley (changes requested):
Small nit around updating docs, but I am trying to be better about keeping all of that up to date :)

PR #974, opened 2025-11-18
:sparkles: Fix gRPC provider to handle typed slice conversion for protobuf marshaling

### LINKED ISSUE #973: gRPC provider should handle typed slice conversion for protobuf marshaling
## Problem

The gRPC provider client fails when provider configs contain typed slices (e.g., `[]string`) because `structpb.NewStruct()` only accepts `[]interface{}` (or `[]any` in Go 1.18+).

**Error:**
```
proto: invalid type: []string
```

**Current workaround:** Clients (kantra, Kai) must manually use `[]interface{}` in provider configs:
```go
providerSpecificConfig["workspaceFolders"] = []interface{}{fmt.Sprintf("file://%s", path)}
providerSpecificConfig["lspServerArgs"] = []interface{}{"--stdio"}
```

This forces callers to know about protobuf marshaling internals, which is an implementation detail that should be hidden.

## Root Cause

In `provider/grpc/provider.go`:
```go
func (g *grpcProvider) Init(ctx context.Context, log logr.Logger, config provider.InitConfig) (...) {
    s, err := structpb.NewStruct(config.ProviderSpecificConfig)  // ← Fails with typed slices
    if err != nil {
        return nil, provider.InitConfig{}, err
    }
```

The protobuf library's `NewValue()` function uses exact type matching in its type switch:
```go
func NewValue(v any) (*Value, error) {
    switch v := v.(type) {
    case []any:  // ← Only matches []interface{}, not []string
        // ...
    default:
        return nil, protoimpl.X.NewError("invalid type: %T", v)
    }
}
```

### Option 1: Convert slices automatically in gRPC provider

Add conversion logic before marshaling to protobuf:

```go
func convertToProtobufCompatible(config map[string]interface{}) map[string]interface{} {
    result := make(map[string]interface{})
    for k, v := range config {
        result[k] = convertValue(v)
    }
    return result
}

func convertValue(v interface{}) interface{} {
    // Use reflection to convert typed slices to []interface{}
    rv := reflect.ValueOf(v)
    if rv.Kind() == reflect.Slice {
        slice := make([]interface{}, rv.Len())
        for i := 0; i < rv.Len(); i++ {
            slice[i] = rv.Index(i).Interface()
        }
        return slice
    }
    return v
}
```

### Option 2: Change API to require `map[string]any`

Update the provider.Config struct to explicitly require `map[string]any` for ProviderSpecificConfig, making the requirement clear at the type level.

## Impact

This affects all clients that create provider configs for gRPC-based providers:
- kantra (hybrid mode, containerized mode)
- Kai
- Any other tools using analyzer-lsp library

## References

- Protobuf source: https://github.com/protocolbuffers/protobuf-go/blob/master/types/known/structpb/struct.pb.go
- Related kantra PR: https://github.com/konveyor/kantra/pull/625
- Related kantra issue: https://github.com/konveyor/kantra/issues/624

tsanders-rh:
## Summary

Fixes #973 by adding automatic conversion of typed slices (e.g., `[]string`, `[]int`) to `[]interface{}` for protobuf compatibility in the gRPC provider.

## Problem

The gRPC provider failed when processing provider configurations with typed slices because `structpb.NewStruct()` only accepts `[]interface{}`. This forced callers to manually convert slices, which:
- Exposed protobuf marshaling internals to users
- Created error-prone manual type juggling
- Impacted multiple projects (kantra, Kai, other analyzer-lsp consumers)

Example that previously failed:
```go
config := map[string]interface{}{
    "includedPaths": []string{"/path1", "/path2"},  // Error!
}
```

## Solution

Added reflection-based conversion logic that:
- Automatically converts typed slices to `[]interface{}` before marshaling
- Handles nested maps and slices recursively
- Optimizes to avoid unnecessary allocations (3.4x faster for already-correct types)
- Maintains full backward compatibility

## Changes

1. **provider/grpc/provider.go** - Added conversion functions and applied in `Init()` method
2. **provider/grpc/dependency_resolver_client.go** - Applied conversion in `GetLocation()` method  
3. **provider/grpc/provider_test.go** - Added 57 comprehensive test cases

## Testing

- ✅ All 57 new tests pass
- ✅ Existing tests continue to pass
- ✅ Integration tests verify `structpb.NewStruct()` compatibility
- ✅ Optimization tests confirm performance improvements

## Performance

- **Impact**: Negligible (<1 microsecond per conversion)
- **Frequency**: Once per provider initialization (not in hot path)
- **Memory**: <8KB even for large configs (100+ paths)
- **Optimization**: 3.4x faster when no conversion needed

## Benefits

- Users can use natural Go types without manual conversion
- Cleaner, more maintainable code
- Reduced error potential
- Better developer experience

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of nested and typed slices/maps when converting configurations for gRPC calls, ensuring proper conversion to protobuf-compatible structures and preventing marshal failures.

* **Tests**
  * Added extensive tests covering conversion logic, nested/mixed structures, nil semantics, direct-failure cases, and performance optimizations to validate correctness and stability.

### DIFF
--- a/provider/grpc/provider.go
+++ b/provider/grpc/provider.go
@@ -64,7 +64,8 @@ func convertValue(value interface{}) interface{} {
 
 	// Check for typed nil (e.g., nil map, nil slice, nil pointer)
 	// This ensures nil maps are marshaled as null instead of empty structs
-	if v.Kind() == reflect.Ptr || v.Kind() == reflect.Map || v.Kind() == reflect.Slice || v.Kind() == reflect.Array {
+	// Note: Arrays are never nil, so we exclude them from this check
+	if v.Kind() == reflect.Ptr || v.Kind() == reflect.Map || v.Kind() == reflect.Slice {
 		if v.IsNil() {
 			return nil
 		}
@@ -132,7 +133,8 @@ func requiresConversion(value interface{}) bool {
 	v := reflect.ValueOf(value)
 
 	// Typed nil values don't need conversion
-	if v.Kind() == reflect.Ptr || v.Kind() == reflect.Map || v.Kind() == reflect.Slice || v.Kind() == reflect.Array {
+	// Note: Arrays are never nil, so we exclude them from this check
+	if v.Kind() == reflect.Ptr || v.Kind() == reflect.Map || v.Kind() == reflect.Slice {
 		if v.IsNil() {
 			return false
 		}
--- a/provider/grpc/provider_test.go
+++ b/provider/grpc/provider_test.go
@@ -363,6 +363,28 @@ func TestRequiresConversion(t *testing.T) {
 	}
 }
 
+// TestArrayConversion verifies that arrays are properly converted
+func TestArrayConversion(t *testing.T) {
+	// Test with a non-nil array type
+	arr := [2]string{"a", "b"}
+
+	result := convertValue(arr)
+
+	// Should be converted to []interface{}
+	slice, ok := result.([]interface{})
+	if !ok {
+		t.Fatalf("Expected []interface{}, got %T", result)
+	}
+
+	if len(slice) != 2 {
+		t.Errorf("Expected length 2, got %d", len(slice))
+	}
+
+	if slice[0] != "a" || slice[1] != "b" {
+		t.Errorf("Expected [a, b], got %v", slice)
+	}
+}
+
 // TestNilMapSemantics verifies the behavior of nil maps in nested structures
 func TestNilMapSemantics(t *testing.T) {
 	// This test documents the nil map behavior difference from structpb.NewStruct


provider/grpc/provider.go
```diff
@@ -38,6 +39,114 @@ type grpcProvider struct {
...
+
+	case reflect.Map:
+		// Maps always need recursive processing
+		return true
+
+	default:
+		// Primitives don't need conversion
+		return false
+	}
+}
```
tsanders-rh: https://github.com/konveyor/analyzer-lsp/pull/974/commits/07ef223fcad4e687ac4604b961246679a3697840

PR #993, opened 2025-11-24
:bug: Use open source index file

### LINKED ISSUE #999: Open source dependencies are not excluded
The Java bundle no longer has access to the open source index file from the Java provider, which means it is unable to exclude open source dependencies during analysis. This causes some rules to time-out, as the search takes much longer than expected.

eemcmullan:
Fixes #999


dymurray (approved):

jmle (approved):

shawn-hurley (approved):
If this fixes the build let's not wait on my comment

PR #1006, opened 2025-11-27
:bug: Fix builtin provider to respect filepath filtering from chained conditions

### LINKED ISSUE #1005: builtin provider does not respect filepath filtering from chained conditions
## Problem Description

The `builtin` provider does not respect filepath filtering from the `from:` clause in chained conditions, unlike the gRPC providers (nodejs, java, dotnet) which were fixed in PR #1004.

## Test Case

When using a chained rule like this:

```yaml
when:
  and:
    - as: tsFiles
      nodejs.referenced:
        pattern: Button
        ignore: true
    - builtin.fileContent:
        filepaths: '{{tsFiles.filepaths}}'
        pattern: isSelectable
        from: tsFiles
```

**Expected Behavior:**
The `builtin.fileContent` condition should only search in the files identified by the `tsFiles` variable from the first condition (respecting the `from: tsFiles` scope).

**Actual Behavior:**
The `builtin.fileContent` provider searches ALL files in the codebase, ignoring the filepath scope passed via `from: tsFiles`.

## Root Cause

The builtin provider does not call `ProviderContext.GetScopedFilepaths()` to extract and apply filepath filtering from chained conditions.

## Proposed Solution

Implement filepath filtering in the builtin provider similar to what was done for gRPC providers in PR #1004:

1. Extract scoped filepaths using `ProviderContext.GetScopedFilepaths()`
2. Build maps for O(1) lookups of included/excluded paths
3. Filter results to only include files within scope
4. Handle URI normalization for cross-platform compatibility

Reference implementation: See changes to `nodejs/service_client.go`, `java/service_client.go`, and `dotnet/service_client.go` in PR #1004.

## Related

- Closes #1002 (partially - gRPC providers fixed, builtin provider remains)
- Related to PR #1004 (fixed gRPC providers)

## Files to Modify

The builtin provider implementation that needs filepath filtering support (likely in the provider's evaluation logic where it processes `filepaths` parameter).

tsanders-rh:
## Summary
Fixes #1005 

The builtin provider did not properly filter files when using chained conditions. This PR adds support for filepath filtering in `builtin.fileContent`, `builtin.json`, and `builtin.xmlPublicID` capabilities.

## Changes

1. **Added `Filepaths` field to `fileContentCondition`** - Allows `builtin.fileContent` to accept filepath scopes from chained conditions
2. **Updated filecontent case** - Now passes `ConditionFilepaths` to `FileSearcher.Search()`
3. **Fixed JSON bug** - Changed from `cond.XML.Filepaths` to `cond.JSON.Filepaths`
4. **Fixed XMLPublicID bug** - Changed from `cond.XML.Filepaths` to `cond.XMLPublicID.Filepaths`
5. **Added `omitempty` tag** - Fixed inconsistency in `xmlPublicIDCondition.Filepaths` for consistent JSON serialization
6. **Added comprehensive tests** - 3 new test cases covering different filepath filtering scenarios

## Example Usage

After this fix, chained conditions like the following now work correctly:

```yaml
when:
  and:
    - as: tsFiles
      nodejs.referenced:
        pattern: Button
        ignore: true
    - builtin.fileContent:
        filepaths: "{{tsFiles.filepaths}}"
        pattern: isSelectable
        from: tsFiles
```

The `builtin.fileContent` condition now correctly searches only the files identified by the `tsFiles` variable instead of searching the entire codebase.

## Test Results

✅ All 30 existing tests pass  
✅ All 3 new filepath filtering tests pass  
✅ Total test runtime: ~31s

## Related Work

This PR complements #1004, which implemented filepath filtering for gRPC providers (Java, .NET, Node.js). The builtin provider now has consistent behavior with those providers.

## Summary by CodeRabbit

* **New Features**
  * Filepath-based filtering added to scope file-content and XML searches more precisely.
  * Chain templates now expose filepath/extras/excludedPaths to templates (all fields available when rendering).

* **Improvements**
  * Filepath parameters are optional in JSON payloads and supported in JSON/YAML templating.
  * Template rendering context updated so chained conditions intersect filepath scopes predictably.

* **Tests**
  * Expanded tests for filepath filtering, include/exclude rules, scoped scenarios.

* **Documentation**
  * Added guidance on filepath filtering behavior in chained conditions.

### DIFF
--- a/engine/conditions.go
+++ b/engine/conditions.go
@@ -316,6 +316,17 @@ type ChainTemplate struct {
 	ExcludedPaths []string               `yaml:"excludedPaths,omitempty" json:"excludedPaths,omitempty"`
 }
 
+// ToMap converts the ChainTemplate to a map suitable for mustache template rendering.
+// All fields are always exposed (even if empty) to ensure templates can reliably
+// reference them without worrying about undefined values.
+func (c *ChainTemplate) ToMap() map[string]interface{} {
+	return map[string]interface{}{
+		"filepaths":     c.Filepaths,
+		"excludedPaths": c.ExcludedPaths,
+		"extras":        c.Extras,
+	}
+}
+
 func (c *ChainTemplate) FilterIncidentsByFilePaths(incidents []IncidentContext) []IncidentContext {
 	//Convert the Filepaths and ExlucedFilePaths back to scopes.
 
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -687,12 +687,7 @@ func templateCondition(condition []byte, ctx map[string]engine.ChainTemplate) ([
 	// reliably reference them without worrying about undefined values
 	yamlCtx := make(map[string]interface{})
 	for key, template := range ctx {
-		templateMap := make(map[string]interface{})
-		// Always set all fields, even if empty, for template consistency
-		templateMap["filepaths"] = template.Filepaths
-		templateMap["excludedPaths"] = template.ExcludedPaths
-		templateMap["extras"] = template.Extras
-		yamlCtx[key] = templateMap
+		yamlCtx[key] = template.ToMap()
 	}
 
 	s, err := mustache.RenderRaw(s, true, yamlCtx)


engine/conditions.go
```diff
@@ -311,9 +311,9 @@ func gatherChain(start ConditionEntry, entries []ConditionEntry) []ConditionEntr
 
 // Chain Templates are used by rules and providers to pass context around during rule execution.
 type ChainTemplate struct {
-	Filepaths     []string               `yaml:"filepaths,omitempty"`
-	Extras        map[string]interface{} `yaml:"extras,omitempty"`
-	ExcludedPaths []string               `yaml:"excludedPaths,omitempty"`
+	Filepaths     []string               `yaml:"filepaths,omitempty" json:"filepaths,omitempty"`
+	Extras        map[string]interface{} `yaml:"extras,omitempty" json:"extras,omitempty"`
+	ExcludedPaths []string               `yaml:"excludedPaths,omitempty" json:"excludedPaths,omitempty"`
```
shawn-hurley: I don't think that this change is correct. 

AFAICT, all providers get the `ProviderContext` not the ChainTemplate. The ChainTemplate is only for engine and the internal ProviderCondition. I don't think this is necessary to fix this bug and may lead to confusion.

provider/internal/builtin/provider.go
```diff
@@ -25,8 +25,9 @@ type builtinCondition struct {
 }
 
 type fileContentCondition struct {
-	FilePattern string `yaml:"filePattern" json:"filePattern,omitempty" title:"FilePattern" description:"Only search in files with names matching this pattern"`
-	Pattern     string `yaml:"pattern" json:"pattern" title:"Pattern" description:"Regex pattern to match in content"`
+	FilePattern string   `yaml:"filePattern" json:"filePattern,omitempty" title:"FilePattern" description:"Only search in files with names matching this pattern"`
+	Pattern     string   `yaml:"pattern" json:"pattern" title:"Pattern" description:"Regex pattern to match in content"`
+	Filepaths   []string `yaml:"filepaths" json:"filepaths,omitempty" title:"Filepaths" description:"Optional list of files to scope down search"`
```
shawn-hurley: Should we make this a helper struct in provider lib so that every referenced condition could just embed it to keep it consistant between everyone?

provider/internal/builtin/service_client_test.go
```diff
@@ -693,6 +693,72 @@ func Test_builtinServiceClient_Evaluate_InclusionExclusion(t *testing.T) {
...
+				Filecontent: fileContentCondition{
+					Pattern: "(fox|app.config.property = .*)",
+					Filepaths: []string{fmt.Sprintf("%s %s",
+						filepath.Join(baseLocation, "dir_a", "a.txt"), filepath.Join(baseLocation, "dir_b", "b.txt")),
+					},
+				},
+			},
+			chainTemplate: engine.ChainTemplate{
+				Filepaths: []string{
+					// here we test whether intersection works as expected
```
shawn-hurley: I am a little confused by this intersection bit here, and think that will need to document this for rule authors. 

IIUC, this would mean that anytime I use a From/As block, then the file paths are automatically set, to the intersection of what I have provided as the rule author and the output of the <As> conditions file paths? 

This seems like a pretty big breaking change, and if there is no way to turn it off, then I think we need to rework it.

provider/provider.go
```diff
@@ -682,7 +682,26 @@ func templateCondition(condition []byte, ctx map[string]engine.ChainTemplate) ([
 	s := strings.ReplaceAll(string(condition), `'{{`, "{{")
 	s = strings.ReplaceAll(s, `}}'`, "}}")
 
-	s, err := mustache.RenderRaw(s, true, ctx)
+	// Handle array template references by converting ctx to have YAML-serialized filepaths
+	yamlCtx := make(map[string]interface{})
```
shawn-hurley: I am also confused on where the go struct is marshaled to yaml here.

engine/conditions.go
tsanders-rh: The ExcludedPaths field itself was already in the codebase (you added it in b062f37 for #786). This commit only added JSON struct tags for consistency.

  What changed:
  // Before (main):
  ExcludedPaths []string `yaml:"excludedPaths,omitempty"`

  // After (this PR):
  ExcludedPaths []string `yaml:"excludedPaths,omitempty" json:"excludedPaths,omitempty"`

  Why it's needed:

  All three ChainTemplate fields (Filepaths, Extras, ExcludedPaths) needed JSON tags added for consistency. Without them, Go's default JSON marshaling capitalizes exported field names:
  - Without tag: {"Filepaths": [...], "ExcludedPaths": [...]}  (capitalized)
  - With tag: {"filepaths": [...], "excludedPaths": [...]}  (matches YAML convention)

  This matters for:

  1. Mustache template rendering - The templateCondition() function in provider/provider.go converts ChainTemplate structs to maps for mustache. Without lowercase JSON tags, template references like {{poms.excludedPaths}} might serialize with capitalized keys.
  2. API consistency - If ChainTemplate is ever serialized to JSON (GRPC, REST APIs, logging), field names should match the YAML convention we use everywhere else.
  3. Future-proofing - Ensures consistent serialization regardless of whether we use YAML or JSON encoding.

  The ExcludedPaths field is actively used in engine/conditions.go:336-346 in the FilterIncidentsByFilePaths method. This change doesn't affect that functionality - it only ensures proper serialization.

  If you'd prefer, we could remove the JSON tags entirely from all three fields, but that would risk inconsistent serialization if ChainTemplate is ever marshaled to JSON.

provider/internal/builtin/service_client_test.go
shawn-hurley: I do think that we need to document this, lets just make sure we add that to the chaining documentation?

tsanders-rh: e1a1251

Issue #910, opened 2025-09-24
Consider using ripgrep tool for files search


aufi:
During experiments with AI agents, I've seen an `rg` command used for searching in larger source files, that might be interesting also for analyzer builtin provider. Not pushing (and don't have any benchmark on analyzer yet), just creating the issue to not forget about it.

More details about the tool: https://github.com/BurntSushi/ripgrep

shawn-hurley:
I agree with the overall goal, we do need to find some way to speed up the file searching, because last i checked it is a large portion of the time taken.

PR #1007, opened 2025-11-28
:sparkles: Draft file content search with ripgrep

aufi:
The aim is to speedup search in file content in builtin provider.

Note, it is needed clarify how to get ripgrep tool downstream (it should be quite common, but default RHEL9 repos don't include it, EPEL does, more about ripgrep https://github.com/BurntSushi/ripgrep).

Related to https://github.com/konveyor/analyzer-lsp/issues/910


aufi:
Closing due to priority/focus change, feel free continue with this if feels relevant.

PR #1033, opened 2025-12-05
:sparkles: Enabling a more ergonomic and controlled entrypoint for the analyzer-lsp library

### LINKED ISSUE #992: Expose helper function(s) to do rule parsing, engine setup, and triggering evaluation
So far, we had been relying on cmd/main.go to setup and run analysis. We have been moving away from using cmd/main.go in our projects and using analyzer APIs directly. As a result, the things we do in the command have been called individually in different projects. We need a set of API defined for these common tasks so the logic is not repeated across Kantra and Kai. While Hub uses cmd/main.go for now, we want to make it use analyzer-lsp as a library instead too. 

Here are some things we commonly do: 

1. Parse rules. Kai also filters rules based on some additional criteria. 
2. Init all providers needed for the rules. While initing these, we also capture any returned initConfigs for builtin provider and add those to list of providers. 
3. Finally, we init all builtin providers (including any additional ones returned by a language provider). We also make sure each location has a default builtin provider even if it is not specified
3. Call Prepare() on all providers.
4. Prepare any additional options for the rules engine including Progress Reporter, scopes etc. 
5. Call RunRulesWithOptions()...internally calls Evaluate() on all rules.
5. Stop() the provider.

shawn-hurley:
fixes #992

## Summary by CodeRabbit

* **New Features**
  * New analyzer API with CLI-driven configuration, unified progress system, and dependency export support.

* **Documentation**
  * Added package docs and updated progress-reporting guidance and examples.

* **Bug Fixes**
  * Improved error aggregation and more robust dependency/decompilation reporting.

* **Refactor**
  * Major rework of progress collectors/reporters and provider initialization lifecycle.

* **Tests**
  * Expanded unit and integration tests across analyzer, progress, and providers.

* **Other**
  * Demo/example paths and provider configs updated for new layout.


shawn-hurley (commented):
## Code Rabbit Review Resolution

I've addressed the following Code Rabbit review comments:

### Fixed Issues:
✅ **cmd/analyzer/main.go** - Added error handling for progress.New() before using it
✅ **cmd/analyzer/main.go** - Fixed ParseRules() to exit on error instead of just logging
✅ **cmd/analyzer/main.go** - Fixed ProviderStart() to exit on error instead of just logging  
✅ **cmd/analyzer/main.go** - Added error handling for GetDependencies()
✅ **cmd/analyzer/main.go** - Fixed comment from '+5' to '+4' to match actual code
✅ **cmd/dep/main.go** - Added error handling for progress.New()
✅ **konveyor/analyzer_test.go** - Fixed GetProviders() to return all providers when no filters provided
✅ **konveyor/util.go** - Fixed duplicate builtin config by moving append outside loop
✅ **konveyor/util_test.go** - Updated test expectations for fixed behavior
✅ **lsp/base_service_client/cmd_dialer.go** - Fixed zombie process leak by calling Cmd.Wait()
✅ **progress/collector/throttled_collector.go** - Fixed panic recovery to use defer unlock properly
✅ **progress/reporter/channel_reporter.go** - Removed debug print statement
✅ **provider/server.go** - Added error logging for progress.New() failures
✅ **generic_external_provider/provider.go** - Fixed error propagation to return actual error
✅ **docs/konveyor-package.md** - Fixed 'three' to 'four' phases

### Not Applicable:
❌ **konveyor/util.go** (lines 23-30) - Current error handling behavior is intentional and correct
❌ **konveyor/analyzer.go** (lines 150-174) - No data race exists; channel blocking provides proper synchronization

All tests pass including race detector.

cmd/analyzer/main.go
```diff
@@ -5,26 +5,20 @@ import (
...
 	"sort"
-	"strings"
 	"sync"
 
 	logrusr "github.com/bombsimon/logrusr/v3"
 	"github.com/go-logr/logr"
-	"github.com/konveyor/analyzer-lsp/engine"
-	"github.com/konveyor/analyzer-lsp/engine/labels"
-	"github.com/konveyor/analyzer-lsp/output/v1/konveyor"
+	"github.com/konveyor/analyzer-lsp/konveyor"
```
jmle: Not sure about the `konveyor` package name - aren't for instance the `external-providers` part of konveyor?

shawn-hurley: Do you have suggestions on the entrypoint package name? 

I was torn, I didn't want it to be `analyzer.Analayzer` and I really couldn't come up with another good option. 

I agree there are many things that are also in the konveyor, just thinking that the package is already scoped to analyzer-lsp, and this is meant to be the way "konveyor" calls this library. Let me know if that makes any sense or if you have any suggestions.

core/analyzer.go
```diff
@@ -0,0 +1,419 @@
...
+				a.log.Info("provider init", "provider", pv.Name)
+				additionalBuiltins, err := pv.provider.ProviderInit(a.ctx, nil)
+				if err != nil {
+					a.log.Error(err, "unable to init provider")
+					providerInitErrors = append(providerInitErrors, err)
+				}
+				a.collector.Report(progress.Event{
+					Stage:   progress.StageProviderInit,
+					Message: fmt.Sprintf("started provider: %s", pv.Name),
+					Current: i + 1,
```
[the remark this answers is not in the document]
shawn-hurley: Just a heads up, but this is fixed in the Go programming language:

https://go.dev/blog/loopvar-preview#the-problem

konveyor/analyzer.go
```diff
@@ -0,0 +1,419 @@
...
+
+func (a *analyzer) RulesetFilepaths() map[string]string {
+	if len(a.ruleset) == 0 {
+		a.log.Info("no ruleset's to get get labels from")
+		return map[string]string{}
+	}
+
+	filePaths := map[string]string{}
+	// TODO: Need to save more info on analyzer when looping through files.
+	return filePaths
```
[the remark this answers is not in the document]
shawn-hurley: I think that this was meant to be implemented later. I will comment out the interface and this block of code and add a TODO and a follow-up?

Thoughts?

shawn-hurley:
Note: these tests **SHOULD** pass; there should be no need to migrate to the new entrypoint immediately. Going to dig into the tests to figure out what is breaking

konveyor/analyzer.go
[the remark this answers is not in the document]
jmle: Sounds good @shawn-hurley

core/analyzer.go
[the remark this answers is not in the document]
jmle: Nice!

cmd/analyzer/main.go
jmle: Then maybe `api` could make sense?

shawn-hurley: `github.com/konveyor/analyzer-lsp/api` would be the import and then `api.NewAnalyzer` 

Yeah, I think that could make some sense. I am for that. @jortel mentioned something about this. I am wondering what his thoughts would be?

jortel: I understand not wanting analyzer.Analyzer and would try to avoid it too.

Given the scope of `api/` at the root of the project, it suggests that it provides all of the public interfaces , types, consts (and I assume it does not).

Perhaps `core` instead of `api`.   Reads core.NewAnalyzer() (or potentially just New())  which returns a `core.Analyzer`.

shawn-hurley: I could be convinced of either of those package names. I think I lean towards core right now. 

Something about "API" just makes me think "web API"; I don't know why, but it does. 

@jmle thoughts on `core`?

provider/grpc/provider.go
```diff
@@ -358,12 +362,21 @@ func (g *grpcProvider) Init(ctx context.Context, log logr.Logger, config provide
 	additionalBuiltinConfig := provider.InitConfig{}
 	if r.BuiltinConfig != nil {
 		additionalBuiltinConfig.Location = r.BuiltinConfig.Location
+		additionalBuiltinConfig.DependencyPath = r.BuiltinConfig.DependencyPath
 	}
+
+	// For right now, the only progress reported by a init config will be the
+	// provider prepare
+	collector := collector.NewThrottledCollector(progress.StageProviderPrepare)
+	g.progress.Subscribe(collector)
```
jmle, line 371: Do we ever `Unsubscribe()` or cancel the context for this at some point?

core/analyzer.go
```diff
@@ -0,0 +1,480 @@
...
+
+var _ Analyzer = &analyzer{}
+
+// ParseRules
+// Either use the the rules that are set during anlaysis creation time, or use the rules that you pass here.
+func (a *analyzer) ParseRules(rulePaths ...string) (Rules, error) {
+	a.log.Info("Parsing rules")
+	// Now we have all the provider clients that have been configured. We can look at the rules to determine which are needed.
+	collector := collector.New()
+	a.progress.Subscribe(collector)
```
jmle, line 59: Same here as in the gRPC client, should we unsubscribe at some point to avoid potential leaks?

core/analyzer.go
```diff
@@ -0,0 +1,480 @@
...
+	// TODO: Handle Scopes
+	ruleset := a.engine.RunRulesWithOptions(a.ctx, a.ruleset, []engine.RunOption{
+		engine.WithProgressReporter(engineOptions.progressReporter),
+	}, engineOptions.selectors...)
+
+	sort.SliceStable(ruleset, func(i, j int) bool {
+		return ruleset[i].Name < ruleset[j].Name
+	})
+	a.log.Info("finished running analysis", "rulesets", ruleset)
+	return ruleset
```
jmle, line 316: Could it be worth returning an error as part of this function?

cmd/analyzer/main.go
```diff
@@ -79,345 +68,130 @@ func AnalysisCmd() *cobra.Command {
...
 			}
+			analyzerCollector := collector.New()
+			analyzerProgress, err := progress.New(
+				progress.WithCollectors(analyzerCollector),
+				progress.WithReporters(progressReporter),
+				progress.WithContext(ctx),
+			)
 
-			tracerOptions := tracing.Options{
-				EnableJaeger:   enableJaeger,
```
jmle, line 115: If we are not using these anymore, we probably should remove them from the CLI options.

jmle (approved):
Looks good, it's a huge PR so most of these last comments come from Claude. The only important comments are the ones related to the potential leaks, what do you think?

Edit: this will break Kantra among other things, right? Before merging, we should probably be sure that we can allocate time to fix everything that is affected asap

cmd/analyzer/main.go
shawn-hurley: I think we need to convert those to follow-ups. I have TODOs in the code, but was planning on adding issues and working on Jaeger and the encoding bits as follow-ups. 

Those would be release blockers though because it is a regression.

Are you ok with that or would you want me to implement now?

core/analyzer.go
shawn-hurley: My first inclination is to agree but, the errors are captured in the ruleset, and I think that is correct because it gives you the error at the ruleset level rather than bunching them all into one error wrapper. 

I wonder if we can find a need for it as we integrate all the callers, and it will become apparent if we need it or not? Thoughts?

provider/grpc/provider.go
shawn-hurley: Only on shutdown right now or stop. I could see a world where Prepare will unsubscribe. I think that is a good idea. 

The concern is a having a list of collectors that will never send anything and we will just increase the memory footprint IIUC?

core/analyzer.go
jmle: That's cool, we can leave them at the rulesets level, and, as you say, if we find the need then we can add it

cmd/analyzer/main.go
jmle: That's totally fine

core/analyzer.go
shawn-hurley: This shoud probably be unsubscribed as a ddefer you are correct. 

I will add this for both Prepare and here, good catch!

shawn-hurley: updated

provider/grpc/provider.go
shawn-hurley: updated good catch!

PR #1034, opened 2025-12-05
:bug: The hardcoded  makes it impossible to set the image correctly 

shawn-hurley:
One should be able to use a full pull spec such as `quay.io/konvelyor/<image>:<tag>` as an IMG env var and it should be used.

## Summary by CodeRabbit

* **Chores**
  * Updated container image references to use localhost-prefixed naming conventions for local development and testing.


savitharaghunathan (approved):
thank you :)

Issue #1032, opened 2025-12-05
Implement logLevel option to provider_settings.json


aufi:
There is a need to have control over log levels on analyzer and providers (primary java provider to control jdtls log verbosity to allow optional full logging output for debugging).

An option `logLevel` should be added to provider_settings entries with each provider and analyzer and its providers including external-providers should respect it.

If logLevel was specified as command options, it will take priority over the logLevel from settings file.

Related to https://github.com/konveyor/tackle2-addon-analyzer/issues/177 and it part of the plan to surface jdtls logs to UI via Hub&addon.

Preffered solution: Use existing addon Data.Verbosity field to tell analyzer (and its providers) expected log level, that will be used by java-external-provider to decide when jdtls should be sent to its stdout.

Steps:
1. analyzer-lsp - update provider_config.json files with optional logLevel field (CLI args would take priority if both were specified), or push back if there is not suitable solution on analyzer side 
2. java-external-provider - update logging of jdtls command execution according to logLevel to optionally provide full jdtls log 
3. addon-analyzer - pass data.Verbosity to generated initConfig (if >0; note, analyzer uses 5 by default "normal" log)
4. decide if this update is also needed for containerized (run-local=false) kantra

PR #1038, opened 2025-12-09
:sparkles: Add logLevel to provider settings

aufi:
There is a logLevel CLI option for analyzer, we need control it also for
providers (including external, like java and its jdtls log that is
exhaustive, but sometimes needed for debugging).

Adding logLevel option to provider settings file (that could be
overwritten by CLI arg).

Related to https://github.com/konveyor/analyzer-lsp/issues/1032

## Summary by CodeRabbit

* **New Features**
  * Added a configurable --log-level startup flag for external providers to control logging verbosity.
  * Added an optional logLevel setting in provider configs so providers can be started with a specific verbosity.
  * Configured providers now propagate and honor the configured log level across launch paths.
  * Provider init emits an informational log of the chosen log level.
  * Backward compatible with existing default log levels.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -191,6 +191,9 @@ type InitConfig struct {
 	// Given a pipe name for the init config, we will use that pipe and connect to an already inited provider.
 	Initialized bool `yaml:"initialized" json:"initialized"`
 
+	// LogLevel for the provider logger verbosity
+	LogLevel *int `yaml:"logLevel,omitempty" json:"logLevel,omitempty"`
+
 	// PrepareProgressReporter is an optional interface for reporting progress during Prepare() phase.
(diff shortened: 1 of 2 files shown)


shawn-hurley (changes requested):
I think that we may be missing a case adding this as a config on the  Provider. 

There are cases where the provider is already running, and setting the log level via flag to the binary will not reflect the log level. I think to fix this, you can change the provider/server.go to handle setting the log level for the resulting loggers of that provider/service_client to use the config value. 

I think we also need to use the value and flag when we run the command. This is should be in provider/provider.go. 

If you need help tracking it all down let me know :)

shawn-hurley:
@aufi I think ou may need to rebase main to get CI to pass for you

PR #1047, opened 2025-12-18
:bug: Fix builtin provider to support multiline pattern matching

### LINKED ISSUE #1043: Builtin Provider: Cannot Match Patterns Across Newlines on Linux/macOS
# Builtin Provider: grep Cannot Match Patterns Across Newlines on Linux

## Problem Description

The builtin provider's filecontent search uses `grep -P` on Linux, which processes files line-by-line and cannot match regex patterns that span across multiple lines. This causes rules with patterns like `<Masthead[^>]*>[^<]*<MastheadToggle` to fail when the matching elements are on different lines in the source code.

## Environment

- **OS**: Linux (Docker container)
- **Component**: analyzer-lsp builtin provider
- **Affected Code**: `provider/internal/builtin/service_client.go:529`

## Current Behavior

On Linux, the builtin provider uses:
```bash
grep -o -n --with-filename -P -- pattern files...
```

This grep command processes the file line-by-line. When a pattern uses character classes like `[^<]*` to match across elements, grep cannot find matches if those elements span multiple lines.

## Expected Behavior

Patterns should match across newlines, similar to how the builtin provider works on Windows (which uses `regexp2.Multiline`).

### Rule Example
```yaml
- ruleID: patternfly-v5-to-patternfly-v6-component-structure-00000
  when:
    and:
    - builtin.filecontent:
        pattern: <Masthead[^>]*>[^<]*<MastheadToggle|<Masthead[^>]*>[^<]*<MastheadBrand
        filePattern: \.(j|t)sx?$
```

### Test Code (Multi-line JSX - Real World Code)
```tsx
// This SHOULD match but DOESN'T on Linux
<Masthead>
  <MastheadToggle>Foo</MastheadToggle>
  <MastheadBrand>Bar</MastheadBrand>
</Masthead>
```

### Test Code (Single-line JSX - Workaround)
```tsx
// This DOES match, but is not realistic production code
<Masthead><MastheadToggle>Foo</MastheadToggle><MastheadBrand>Bar</MastheadBrand></Masthead>
```

## Analysis Log Evidence

From `analysis.log`:
```
time="2025-12-18T06:19:29Z" level=unknown msg="using direct grep (fast path)" pattern="<Masthead[^>]*>[^<]*<MastheadToggle|<Masthead[^>]*>[^<]*<MastheadBrand" totalFiles=1
time="2025-12-18T06:19:29Z" level=unknown msg="no incidents found" ruleID=patternfly-v5-to-patternfly-v6-component-structure-00000
```

The pattern should have matched the JSX code, but grep found 0 incidents because the tags are on separate lines.

## Root Cause

In `service_client.go:performFileContentSearch()`:

- **Windows** (line 500-512): Uses `regexp2.Compile(trimmedPattern, regexp2.Multiline)` ✓ WORKS
- **Linux** (line 524-544): Uses `grep -P` ✗ FAILS for multiline patterns

The grep command lacks multiline support flags.

### Option 1: Use pcregrep with -M flag (Recommended)

Replace `grep -P` with `pcregrep -M`:
```go
args := []string{"-o", "-n", "-M", "--", pattern}
cmd := exec.Command("pcregrep", args...)
```

The `-M` flag enables multiline mode where patterns can match across newlines.

**Pros**: Clean, well-supported tool
**Cons**: Requires pcregrep to be installed in containers

### Option 2: Use parallelWalk on Linux (like Windows)

Remove the platform check and use the `regexp2.Multiline` approach on all platforms:
```go
trimmedPattern := strings.Trim(pattern, "\"")
patternRegex, err := regexp2.Compile(trimmedPattern, regexp2.Multiline)
if err != nil {
    return nil, fmt.Errorf("could not compile provided regex pattern '%s': %v", pattern, err)
}
matches, err := b.parallelWalk(locations, patternRegex)
```

**Pros**: No external dependency, consistent across platforms
**Cons**: May be slower than grep for large projects (though the code already has this path for Windows)

### Option 3: Use grep with null-separated input

Use `grep -z` which treats null bytes as line separators:
```go
args := []string{"-o", "-z", "-n", "--with-filename", "-P", "--", pattern}
```

**Pros**: Uses standard grep
**Cons**: May have other side effects; less tested

## Impact

This issue affects any rule that needs to match JSX/TSX elements spanning multiple lines, which is standard formatting in modern JavaScript/TypeScript codebases. Without this fix, rules must either:
1. Force developers to write unrealistic single-line code
2. Be written in a way that doesn't properly detect the deprecated patterns

## Testing

After implementing the fix, the following test should pass:

**File**: `src/App.tsx`
```tsx
import React from 'react';
import { Masthead, MastheadToggle, MastheadBrand } from '@patternfly/react-core';

const App: React.FC = () => {
  return (
    <div>
      {/* Rule patternfly-v5-to-patternfly-v6-component-structure-00000 */}
      <Masthead>
        <MastheadToggle>Foo</MastheadToggle>
        <MastheadBrand>Bar</MastheadBrand>
      </Masthead>
    </div>
  );
};

export default App;
```

**Expected**: The rule should find 1 incident at the `<Masthead>` element.

## Additional Context

Related files:
- `provider/internal/builtin/service_client.go:497-663` (performFileContentSearch, parallelWalk)
- `provider/lib.go:372` (MultilineGrep - already exists for other use cases)

The `MultilineGrep` function in `lib.go` already supports multiline patterns using `(?s)` flag, but it's only used for `getLocation()`, not for the initial pattern search.

tsanders-rh:
## Summary

Fixes the builtin provider's inability to match regex patterns across newlines on Linux and macOS. This allows rules to properly detect patterns in multiline code, which is standard formatting in modern codebases.

## Problem

The builtin provider used platform-specific implementations:
- **Linux**: `grep -P` (processes line-by-line)
- **macOS**: `perl -ne` (processes line-by-line)  
- **Windows**: `regexp2.Multiline` (supports multiline matching)

This caused patterns like `<Masthead[^>]*>[^<]*<MastheadToggle` to fail when matching across newlines, even though the pattern exists in the file. This is a critical issue for analyzing formatted JSX/TSX and other structured code.

### Example that was broken:
```tsx
// This SHOULD match but DIDN'T on Linux/macOS
<Masthead>
  <MastheadToggle>Foo</MastheadToggle>
</Masthead>
```

## Solution

Unified the implementation to use `regexp2.Multiline` on all platforms (following the Windows approach):
- ✅ Removed platform-specific grep/perl code (~200 lines)
- ✅ Single code path for all platforms
- ✅ No external dependencies
- ✅ Fixed `processFile()` to search entire file content instead of line-by-line
- ✅ Removed chunked pre-check that broke multiline patterns

## Changes

**Modified:**
- `provider/internal/builtin/service_client.go`
  - Removed `performFileContentSearch()` platform conditionals
  - Simplified `processFile()` to search full content
  - Removed `isSlashEscaped()` and `parseGrepOutputForFileContent()` helpers
  - Removed unused imports: `bufio`, `math`, `os/exec`, `runtime`, `strconv`

**Added:**
- `provider/internal/builtin/service_client_test.go`
  - `Test_builtinServiceClient_performFileContentSearch_Multiline` with 3 test cases
- `provider/internal/builtin/testdata/multiline-test.tsx`
  - Test data covering multiline and single-line patterns

### New Tests
```
✅ Test_builtinServiceClient_performFileContentSearch_Multiline
  ✅ multiline pattern matches across newlines
  ✅ multiline pattern with alternative matches MastheadBrand
  ✅ combined pattern with OR
```

### Regression Testing
All 30+ existing tests continue to pass:
- `Test_builtinServiceClient_getLocation`
- `Test_builtinServiceClient_Evaluate_InclusionExclusion` (30 test cases)
- `Test_builtinServiceClient_Evaluate_ExcludeDirs` (7 test cases)

### Manual Verification
Tested with real-world example from issue:
```yaml
pattern: <Masthead[^>]*>[^<]*<MastheadToggle|<Masthead[^>]*>[^<]*<MastheadBrand
```
Now correctly finds 3 matches (2 multiline, 1 single-line) in test file.

## Impact

**Positive:**
- Fixes critical bug preventing multiline pattern matching
- Simpler codebase (net -86 lines)
- Consistent behavior across all platforms
- No external dependencies

**Performance:**
- May be slightly slower on files with no matches (no chunked pre-check)
- Faster on files with matches (already parallelized via `parallelWalk`)
- Correctness prioritized over micro-optimization

## Checklist

- [x] Code follows project style
- [x] All tests pass
- [x] New tests added for multiline functionality  
- [x] No breaking changes
- [x] Documentation unchanged (no limitations to remove)
- [x] Commit signed with DCO

Closes #1043

## Summary by CodeRabbit

* **New Features**
  * Multiline pattern matching for file-content search with a dedicated multiline workflow.

* **Improvements**
  * Faster concurrent per-file processing that distinguishes literal vs. regex patterns.
  * Windows parity for path and line handling, UTF-16-aware offsets, encoding fallbacks, and clearer error messages.
  * File-size safeguards to avoid out-of-memory issues.

* **Tests**
  * Added test data and tests validating multiline matches and exact line numbers.

### DIFF
--- a/provider/internal/builtin/service_client.go
+++ b/provider/internal/builtin/service_client.go
@@ -15,7 +15,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"unicode/utf8"
 
 	"github.com/antchfx/jsonquery"
 	"github.com/antchfx/xmlquery"
@@ -678,8 +677,8 @@ func (b *builtinServiceClient) processFileWithLiteralCheck(path string, regex *r
 			}
 			lastPos = matchStart
 
-			// Calculate UTF-16 code units from start of line to match position
-			charNumber := utf16CodeUnits(content[lineStart:matchStart])
+			// Character position as byte offset from start of line
+			charNumber := matchStart - lineStart
 
 			r = append(r, fileSearchResult{
 				positionParams: protocol.TextDocumentPositionParams{
@@ -767,8 +766,8 @@ func (b *builtinServiceClient) processFileWithLiteralCheck(path string, regex *r
 			}
 			lastPos = matchStart
 
-			// Calculate UTF-16 code units from start of line to match position
-			charNumber := utf16CodeUnits(content[lineStart:matchStart])
+			// Character position as byte offset from start of line
+			charNumber := matchStart - lineStart
 
 			r = append(r, fileSearchResult{
 				positionParams: protocol.TextDocumentPositionParams{
@@ -808,8 +807,8 @@ func (b *builtinServiceClient) processFileWithLiteralCheck(path string, regex *r
 			}
 			lastPos = match.Index
 
-			// Calculate UTF-16 code units from start of line to match position
-			charNumber := utf16CodeUnits([]byte(contentStr[lineStart:match.Index]))
+			// Character position as byte offset from start of line
+			charNumber := match.Index - lineStart
 
 			r = append(r, fileSearchResult{
 				positionParams: protocol.TextDocumentPositionParams{
@@ -1109,9 +1108,6 @@ func (b *builtinServiceClient) processFileWindows(path string, regex *regexp2.Re
 			return nil, err
 		}
 		if match != nil {
-			// Calculate UTF-16 code units from start of line to match position
-			charNumber := utf16CodeUnits([]byte(line[:match.Index]))
-
 			r = append(r, fileSearchResult{
 				positionParams: protocol.TextDocumentPositionParams{
 					TextDocument: protocol.TextDocumentIdentifier{
@@ -1119,7 +1115,7 @@ func (b *builtinServiceClient) processFileWindows(path string, regex *regexp2.Re
 					},
 					Position: protocol.Position{
 						Line:      lineNumber,
-						Character: uint32(charNumber),
+						Character: uint32(match.Index),
 					},
 				},
 				match: match.String(),
@@ -1133,31 +1129,6 @@ func (b *builtinServiceClient) processFileWindows(path string, regex *regexp2.Re
 	}
 
 	return r, nil
-}
-
-// utf16CodeUnits returns the number of UTF-16 code units in the given byte slice.
-// This is required for LSP Position.Character which must be measured in UTF-16 code units.
-// Most Unicode characters (< U+10000) use 1 code unit, while supplementary characters
-// (>= U+10000) use 2 code units (surrogate pairs).
-func utf16CodeUnits(b []byte) int {
-	count := 0
-	for len(b) > 0 {
-		r, size := utf8.DecodeRune(b)
-		if r == utf8.RuneError {
-			// Invalid UTF-8, count as 1
-			count++
-			b = b[1:]
-		} else {
-			// Runes >= 0x10000 require surrogate pairs in UTF-16 (2 code units)
-			if r >= 0x10000 {
-				count += 2
-			} else {
-				count++
-			}
-			b = b[size:]
-		}
-	}
-	return count
 }
 
 func isSlashEscaped(str string) bool {


tsanders-rh:
## Critical Optimization: Skip regexp2 Entirely for Simple Patterns

Previous versions always compiled both standard regexp AND regexp2, keeping regexp2 overhead even for simple patterns.

**New approach:**
1. Try `regexp.Compile()` first
2. **Only** if it fails (regexp2-specific syntax), compile `regexp2.Compile()`

**For "Apache License 1.1" (benchmark):**
- Standard regexp compiles successfully
- regexp2 is **never compiled** or used
- Zero regexp2 overhead

**For complex patterns (issue #1043):**
- Standard regexp compile fails
- Falls back to regexp2 with multiline support
- Chunked pre-check for memory efficiency

This completely eliminates regexp2 overhead for the ~90% of patterns that work with standard regexp, while still supporting multiline matching for the patterns that need it.

provider/internal/builtin/service_client.go
```diff
@@ -495,20 +495,332 @@ type fileSearchResult struct {
...
+	eg.SetLimit(20)
+
+	for _, filePath := range paths {
+		eg.Go(func() error {
+			pos, err := b.processFileWithLiteralCheck(filePath, regex, stdRegex, literalPattern, isLiteral)
+			if err != nil {
+				return err
+			}
+
+			positionsMu.Lock()
```
shawn-hurley: If pos is empty, we should exit early and not take the lock.

provider/internal/builtin/service_client.go
```diff
@@ -737,53 +1051,48 @@ func (b *builtinServiceClient) processFile(path string, regex *regexp2.Regexp) (
...
+	absPath, err := filepath.Abs(path)
+	if err != nil {
+		return nil, err
+	}
 
+	reader = bytes.NewReader(content)
+	r := []fileSearchResult{}
 	scanner := bufio.NewScanner(reader)
-	lineNumber := 1
+	lineNumber := 0
```
shawn-hurley: Why did we change this from 1 to 0? IIRC, this is 1 to make the line number 1-based rather than 0-based, which makes the code snippet display more accurateff.

provider/internal/builtin/service_client.go
tsanders-rh: cea14bfb6c6ec797bbc3ae89b170d161948f12f7

tsanders-rh:
@shawn-hurley - I've put back to 1-based line numbering which violates the lsp-spec, so I guess ignore the code rabbit screaming.  -Todd

provider/internal/builtin/service_client.go
```diff
@@ -737,58 +1083,83 @@ func (b *builtinServiceClient) processFile(path string, regex *regexp2.Regexp) (
...
 	}
 
 	return r, nil
 }
 
+// utf16CodeUnits returns the number of UTF-16 code units in the given byte slice.
+// This is required for LSP Position.Character which must be measured in UTF-16 code units.
+// Most Unicode characters (< U+10000) use 1 code unit, while supplementary characters
+// (>= U+10000) use 2 code units (surrogate pairs).
+func utf16CodeUnits(b []byte) int {
```
shawn-hurley: I am pretty sure that the response for this is for the analyzer, not LSP. We don't need to conform to the LSP spec for this function. 

The position may be put into an LSP spec object, but will only be used by the analyzer. We don't have a defined standard spec for that field, and maybe this is a good time to make that the case. 

I think there are two paths forward:

1. Remove this, and use the starting position in the line for the character. Let's create an issue and make everything conform to the same way of representing the character.
2. We can keep this, and the character will continue to have different values, potentially based on the provider. We should still create an issue to address this across all providers.

provider/internal/builtin/service_client.go
```diff
@@ -692,9 +1027,20 @@ func (b *builtinServiceClient) parallelWalk(paths []string, regex *regexp2.Regex
...
-func (b *builtinServiceClient) processFile(path string, regex *regexp2.Regexp) ([]fileSearchResult, error) {
+// processFileWindows processes a file on Windows using line-by-line scanning
+func (b *builtinServiceClient) processFileWindows(path string, regex *regexp2.Regexp) ([]fileSearchResult, error) {
+	// Check file size before loading to prevent OOM on large files (e.g., JARs)
+	const maxFileSize = 100 * 1024 * 1024 // 100MB limit
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if fileInfo.Size() > maxFileSize {
```
shawn-hurley: So I think we have two things now to deal with large files:

1. If a file is larger than 100MB, we skip it.
2. If it is smaller than 100MB then we load 1MB chunks into memory w/ 8KB of overlap to handle cross chunk matches. 

I just wanted to make sure this is correct and that I am following.

So IIRC, when we started writing the golang-based approach, we initially decided not to have a file-size cap and instead always load chunks. If the file is larger than 100mb then we are talking about a one million lines of code (give or take). 

I am personally ok with removing the chunking and using 100MB as a cutoff, but want to open up to @pranavgaikwad, who I believe advocated for the chunking. If I am misremembering, I am very sorry for the back-and-forth, especially if this was in the initial PR.

shawn-hurley (changes requested):

tsanders-rh:
@shawn-hurley - I think the challenges with a strictly standard grep approach is that multi-line features:
 - Change the output format in ways that break line number detection
  - Aren't universally available
  - Don't work on Windows

This is why we are taking the approach of:
  - Simple patterns → grep/perl (line-by-line)
  - Multiline/complex patterns → Go implementation

provider/internal/builtin/service_client.go
tsanders-rh: Let me clarify the current state:

  Chunking existed before this PR:

  The original code on main already had chunking, but it worked the same way - it loaded the entire file into memory with os.ReadFile() first, then processed it in 1MB chunks:
```
  // Original code on main:
  content, err = os.ReadFile(path)  // Load entire file
  reader := bytes.NewReader(content)
  buffer := make([]byte, 1024*1024)  // Then chunk through it
```
  This PR keeps the same pattern:
  1. Load entire file with os.ReadFile()
  2. Process in chunks (1MB with 8KB overlap for multiline patterns)

  The chunking in both cases reduces string allocations during matching, but doesn't reduce peak memory usage since the entire file is already in memory.

  What this PR added:

  The 100MB file size guard is actually new in this PR and an improvement over the original code which had no size limit. It prevents the worst-case scenario of loading massive JAR files.

  Regarding the memory concern:

  Pranav's math is correct - with 20 parallel goroutines, worst case could be ~2GB (20 × 99MB) which exceeds our headroom. A few options:

  1. Lower the file size limit (e.g., 50MB → max 1GB across 20 files)
  2. Reduce parallelism (lower errgroup.SetLimit from 20)
  3. Implement true streaming (read file in chunks without loading entirely)

  The memory issue existed before this PR; we just haven't addressed it yet. Should we tackle it in this PR or file a follow-up issue?

PR #1075, opened 2026-01-23
:sparkles: Initial implementation of mcp server

JonahSussman:
This PR needs to be cleaned up a lot before merging in, but I wanted to get an initial draft using AI assistance to see what's possible.


shawn-hurley:
@JonahSussman Between that and the enhancement, I would focus on getting agreement and reviews on the enhancement. If you want to do both, I wouldn't be upset. 

On that PR, I just wanted to go through all of code rabbits comments to check stuff, if you wanted to start working on them and submit PR's against the branch in my repo that would be a good way to do it

shawn-hurley (changes requested):
I think that the entrypoint branch is at a place to start experimenting with integration if you would like to do that. 

Hopefully you can get that working so that we know if there are things missing from that, that you would need.

Issue #1083, opened 2026-01-29
[BUG] filePattern parameter finds zero files when repository path contains substring matching exclude patterns


JonahSussman:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

Latest main branch (commit: fe83a80)

### Priority

Major

### Current Behavior

When using `builtin.filecontent` with the `filePattern` parameter, the analyzer sometimes finds zero files even when matching files exist in the repository.

This occurs when the repository path contains substrings that match default exclude directory patterns. For example, running the analyzer somewhere in a directory prefixed with `/home/user/Projects/github.com/...` and in a directory without `.git` can trigger the bug because `.git` (in the default excludes) gets treated as a regex and matches `/git` in the path.

### Expected Behavior

The `filePattern` parameter should find all files matching the pattern, regardless of the repository's absolute path. Exclude patterns like `.git` should only exclude literal `.git` directories, not paths containing the substring `/git`.

### How Reproducible

Always (Default)

### Steps To Reproduce

Run the following bash script:

```bash
#!/bin/bash
set -e

# Configuration
BRANCH="main"
REPO_DIR="path/to/repo"
TEST_DIR=$(mktemp -d -t analyzer-test-XXXXXX)

echo "=========================================="
echo "REPO_DIR: $REPO_DIR"
echo "BRANCH:   $BRANCH"
echo "TEST_DIR: $TEST_DIR"
echo "=========================================="
echo

# Navigate to repo and checkout branch
cd "$REPO_DIR"
echo "[1/6] Checking out branch: $BRANCH"
git checkout "$BRANCH" 2>&1 # | grep -v "^Already on"

# Build analyzer
echo "[2/6] Building analyzer..."
go build -o "$TEST_DIR/analyzer" ./cmd/analyzer 2>&1 | tail -5

# Create test directory with 'git' in name (triggers bug)
echo "[3/6] Creating test files..."
mkdir -p "$TEST_DIR/test-dir-with-git-in-name"

cat > "$TEST_DIR/test-dir-with-git-in-name/test.jsx" << 'JSXEOF'
import { Page, Masthead } from '@patternfly/react-core';

function MyComponent() {
  return (
    <Page header={<Masthead />} />
  );
}
JSXEOF

# Create test rule
cat > "$TEST_DIR/test-rule.yaml" << 'YAMLEOF'
---
- ruleID: test-filecontent-pattern
  labels:
  - konveyor-analysis
[...]

PR #1084, opened 2026-01-29
:bug: Fix filePattern parameter finds zero files when repository path contains substring matching exclude patterns

JonahSussman:
Fixes: #1083 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved literal pattern handling for dot-prefixed directories and virtualenvs.
  * Fixed inconsistent Windows path normalization to ensure reliable file matching and exclusions.

* **Tests**
  * Expanded test coverage for rule parsing, AND-condition evaluation with combined file-content checks, file searching behavior, dependency and error scenarios, and many parsing fixtures.

* **Chores**
  * Added numerous test data examples and sample files to exercise edge cases.

### DIFF
--- a/provider/internal/builtin/wcmanger.go
+++ b/provider/internal/builtin/wcmanger.go
@@ -95,44 +95,41 @@ func (t *workingCopyManager) startWorker() {
 			if !ok {
 				return
 			}
-			// Keep original path for filePath (used for exclusion matching)
-			originalPath := change.Path
-			// Strip volume label on windows for wcPath creation only
-			pathForWc := change.Path
+			// we need to get rid of volume label on windows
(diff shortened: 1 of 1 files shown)


shawn-hurley (changes requested):
If I understand correctly, this means that if a regex is "t.st", we would treat it as "t\.st". 

I feel this violates the user's intent. If the user wants to treat the dot as a literal rather than a regex wildcard, they can escape it. Am I missing something?

The tests for dependency rules (and the no-dependency-rule option), as well as schema creation, are a good addition. Thank you!

Some of the other tests, though, I think should already be covered or are good additions, but should be part of the primary test cases, IMO. Unless I am missing a reason why they can't be (which is very possible).

PR #1115, opened 2026-02-23
:bug: Reset decomp error after setting dummy dep 

eemcmullan:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when processing invalid dependencies during JAR analysis—the system now gracefully continues with placeholder artifacts instead of failing outright.

* **Tests**
  * Added comprehensive unit tests for JAR artifact processing, including validation of error recovery and path generation scenarios.

### DIFF
--- a/external-providers/java-external-provider/pkg/java_external_provider/dependency/jar.go
+++ b/external-providers/java-external-provider/pkg/java_external_provider/dependency/jar.go
@@ -47,6 +47,8 @@ func (j *jarArtifact) Run(ctx context.Context, log logr.Logger) error {
 			Sha1:        "",
 		}
 		dep = newDep
+		// recover with dummy coordinates so decomp can continue
+		err = nil
 	}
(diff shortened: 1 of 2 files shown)


jmle (approved):

external-providers/java-external-provider/pkg/java_external_provider/dependency/jar.go
```diff
@@ -47,6 +47,8 @@ func (j *jarArtifact) Run(ctx context.Context, log logr.Logger) error {
 			Sha1:        "",
 		}
 		dep = newDep
+		// recover with dummy coordinates so decomp can continue
+		err = nil
```
shawn-hurley: nit, should we do this in the `err != nil {` block on line 35? 

Either way good change, thank you! 

Is there a way to add a test for this?

shawn-hurley (approved):
I am good merging as is, but just some nits

PR #1116, opened 2026-02-23
:bug: Fix dedup index bug and config variable shadowing in provider

JonahSussman:
## Summary

Fixes two bugs in `provider/provider.go`:

- `deduplicateDependencies` stored wrong indices, breaking indirect-to-direct dependency upgrades
- `validateUpdateInternalProviderConfig` silently dropped nested provider configs due to variable shadowing

## Bug 1: Wrong index in `deduplicateDependencies`

The data structure is a map of slices:

```go
deduped := map[uri.URI][]*Dep{}
//         ^outer map    ^inner slice
```

When a dep is appended, its index is stored in `depSeen` so it can be found later (e.g. to upgrade it from indirect to direct). Say we're processing deps for `file:///pom.xml` and we've appended 3 deps:

```
deduped = {
    "file:///pom.xml": [libA, libB, libC]   // inner slice, len = 3
}
// outer map has 1 key, so len(deduped) = 1
```

When libB was appended at index 1, the old code stored its index as:

```go
depSeen[id+"indirect"] = intPtr(len(deduped) - 1)   // = 1 - 1 = 0  (wrong!)
```

It used the length of the outer map (number of URI keys, which is 1) instead of the inner slice (number of deps, which was 2 at the time). So it stored index `0` instead of index `1`.

Later, when a direct version of libB shows up and the code tries to upgrade it:

```go
deduped[uri][*depSeen[id+"indirect"]].Indirect = false
//          ^ this resolves to index 0, which is libA, not libB
```

It modifies libA instead of libB. The fix changes it to `len(deduped[uri]) - 1`, which gives the actual position in the slice:

```go
// before
depSeen[id+"indirect"] = intPtr(len(deduped) - 1)

// after
depSeen[id+"indirect"] = intPtr(len(deduped[uri]) - 1)
```

## Bug 2: Variable shadowing in `validateUpdateInternalProviderConfig`

The function builds up a result map called `new`:

```go
new := map[string]interface{}{}   // this is the result we're building
```

When it encounters a nested map, it needs to recursively process it and store the result. The old code was:

```go
new, err := validateUpdateInternalProviderConfig(o)
new[s] = new
```

The `:=` on the first line declares a brand new local variable also called `new`. From that point on, `new` refers to the local one, not the outer one. So `new[s] = new` inserts the local map into itself — a circular self-reference that does nothing useful. The outer `new` map is never touched, so the nested key just disappears from the result.

The fix uses a different variable name:

```go
nested, err := validateUpdateInternalProviderConfig(o)
new[s] = nested   // "new" here refers to the outer map again
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected dependency deduplication logic to ensure indirect and direct dependencies are tracked accurately.
  * Fixed an issue where nested provider configuration settings were not properly preserved during validation.

### DIFF
--- a/provider/provider_bugfix_test.go
+++ b/provider/provider_bugfix_test.go
@@ -128,25 +128,31 @@ func TestValidateUpdateInternalProviderConfig_NestedMap(t *testing.T) {
 }
 
 // Inlined copy of the old buggy validation to show the failure.
+// The shadowed "new" causes new[s] = new to create a circular reference
+// in the local map, while the outer map never gets the nested key at all.
 func TestValidateUpdateInternalProviderConfig_OldBuggyBehavior(t *testing.T) {
-	buggyValidate := func(old map[interface{}]interface{}) (map[string]interface{}, error) {
+	// Returns both the outer result and the local map from the shadowed scope
+	// so we can inspect the circular reference.
+	buggyValidate := func(old map[interface{}]interface{}) (map[string]interface{}, map[string]interface{}, error) {
+		var captured map[string]interface{}
 		new := map[string]interface{}{}
 		for k, v := range old {
 			s, ok := k.(string)
 			if !ok {
-				return nil, fmt.Errorf("key is not a string")
+				return nil, nil, fmt.Errorf("key is not a string")
 			}
 			if o, ok := v.(map[interface{}]interface{}); ok {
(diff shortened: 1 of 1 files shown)


shawn-hurley:
Thank you for the detailed write-up of the fixes. This looks good to me!

I'll give some time for other reviews and then can merge. Thank you!

Issue #1082, opened 2026-01-29
[BUG] C# analysis succeeded with errors


mguetta1:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

latest

### Priority

Major

### Current Behavior

C# **source-only** mode analysis with a custom rule file succeeded with errors
```
errors:
    - severity: Error
      description: '[Analyzer] technology-usage.javaee-technology-usage-00011: unable to get query info: yaml: line 430: did not find expected '','' or '']'''
```

**Application details:**
- URL: https://github.com/konveyor/c-sharp-analyzer-provider
- Root path: testdata/nerd-dinner

**Custom rule file:**
```
- category: mandatory
  customVariables: []
  description: ASP.NET System.Web namespace is not available in .NET Core
  effort: 1
  labels:
  - konveyor.io/source=dotnet
  - konveyor.io/target=dotnet-core
  links:
  - title: Migrate from ASP.NET to ASP.NET Core
    url: https://learn.microsoft.com/en-us/aspnet/core/migration/proper-to-2x
  - title: System.Web namespace migration
    url: https://learn.microsoft.com/en-us/aspnet/core/migration/mvc
  message: |
    The System.Web namespace is not available in .NET Core.
    This namespace contains types that are specific to ASP.NET web applications and is replaced by ASP.NET Core components.

    Migration actions:
    - Replace System.Web.Mvc with Microsoft.AspNetCore.Mvc
    - Replace System.Web.Http with ASP.NET Core Web API
    - Update controller base classes and routing
    - Migrate to ASP.NET Core middleware pipeline
  ruleID: dotnet-core-system-web-01
  when:
    csharp.referenced:
      location: ALL
      pattern: System.Web.Mvc.*
```

### How Reproducible

Always (Default)

### Environment

```markdown

```

### Anything else?

This error is not reproducible when using another csharp aaplication: https://github.com/konveyor/c-sharp-analyzer-provider/tree/main/testdata/net8-sample

shawn-hurley:
We really shouldn't be running those rules against C #, we really need to consider that a set of default rules is tied to a provider.

mguetta1:
I am able to reproduce this issue with Konveyor latest (main). Reopening

mguetta1:
Verified as fixed with latest `tackle2-addon-analyzer` image (rebuilt on Mar. 10). Closing

PR #1122, opened 2026-03-03
:bug: Quote filepaths when passing them to yaml processing during chaining

jmle:
When passing around filepaths for condition chaining, if any of the files has special characters (for instance, `]`), the generated yaml is not valid. We need to quote the filepaths to prevent this from happening.

Fixes https://github.com/konveyor/analyzer-lsp/issues/1082

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced serialization of file paths in conditions to properly handle special characters, spaces, and quotes.

* **Tests**
  * Added comprehensive test coverage for edge cases involving special characters in file path formatting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 target/
 *.metadata*
 jdt.ls-java-project
+examples/c-sharp-dummy-example/obj/
 examples/java/*.project
 examples/python/.venv
 examples/python/__pycache__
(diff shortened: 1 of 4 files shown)


shawn-hurley:
I think that we should just merge this as is, I created an issue to track the refactor and we can figure out the struct bits there. #1123 

We do need to update to yaml.v4, but I think it behaves quite differently so that is going to be something we need to look into.

PR #1124, opened 2026-03-05
:sparkles: builtin provider with Incident cache 

### LINKED ISSUE #1018: [feature] With the addition of `Prepare` the bulitin provider should only make a single pass through the filesystem
For any builtin condition, that does not use a chain condition, we should be able to search for it's condition in the filesystem, during a single pass through the filesytem. We should then be able to cache these results, and in combination with the NotifyChanges, give back results in O(1) time. 

How to handle chained conditions, will be a follow-up. There may be something clever to do, but the work described here is valuable on it's own.

savitharaghunathan:
## Credits
**Assisted by Claude Code**

---

## Summary by CodeRabbit

* **New Features**
  * Faster, index-backed file search with multi-format (text, XML, JSON) pre-processing, improved match location extraction, and XML 1.1 compatibility; automatic cache refresh when files change.

* **Refactor**
  * Preparation now builds a reusable incident cache and pre-compiles search conditions so evaluations prefer cached results with safe fallbacks to on-demand scanning.

* **Tests**
  * Expanded test suite covering indexing, caching, scope/filtering, chained/fallback behavior, pattern matching, and evaluate scenarios.

### DIFF
--- a/provider/internal/builtin/incident_cache.go
+++ b/provider/internal/builtin/incident_cache.go
@@ -21,6 +21,11 @@ import (
 // during Prepare()'s incident caching. Files larger than this are skipped.
 const maxCacheFileSize = 100 * 1024 * 1024 // 100MB
 
+// numWorkers is the number of goroutines used for both the initial
+// incident cache population and the cache refresh worker pool.
+// TODO: make numWorkers configurable via providerSpecificConfig
+const numWorkers = 5
+
 // cacheRefreshRequest is sent from the working copy manager to the
 // builtinServiceClient after a file change has been written to disk.
 type cacheRefreshRequest struct {
@@ -51,11 +56,8 @@ type parsedCondition struct {
 	jsonXPath           string         // json
 }
 
-// makeIncidentCacheKey builds a canonical cache key for a condition.
-// Keys are derived from the parsed condition fields (Pattern, XPath, etc.),
-// not from raw conditionInfo bytes. This ensures Prepare() and Evaluate()
-// produce identical keys as long as YAML unmarshalling is deterministic,
-// which it is for these simple scalar/map fields.
+// makeIncidentCacheKey builds a canonical cache key from a condition's
+// parsed fields (Pattern, XPath, etc.).
 func makeIncidentCacheKey(cap string, cond builtinCondition) incidentCacheKey {
 	switch cap {
 	case "filecontent":
@@ -122,8 +124,10 @@ func isChainedCondition(cap string, cond builtinCondition) bool {
 }
 
 // tryIncidentCache attempts to look up cached incidents for a capability.
-// Returns the incidents, whether a cache hit occurred, and any error.
-func (b *builtinServiceClient) tryIncidentCache(cap string, cond builtinCondition, patterns []string) ([]provider.IncidentContext, bool, error) {
+// If any scoped file has a pending cache refresh, it waits for the refresh
+// to complete before returning results. Returns the incidents, whether a
+// cache hit occurred, and any error.
+func (b *builtinServiceClient) tryIncidentCache(ctx context.Context, cap string, cond builtinCondition, patterns []string) ([]provider.IncidentContext, bool, error) {
 	if !b.prepared || len(b.allParsedConditions) == 0 || isChainedCondition(cap, cond) {
 		return nil, false, nil
 	}
@@ -134,10 +138,17 @@ func (b *builtinServiceClient) tryIncidentCache(cap string, cond builtinConditio
 	if err != nil {
 		return nil, false, err
 	}
-	// If any scoped file is pending cache refresh, fall back to
-	// filesystem walk so Evaluate doesn't return stale/incomplete results.
+	// Wait for any pending cache refreshes to complete before reading
+	// from the cache, so Evaluate never returns stale/incomplete results.
 	for _, f := range scopedFiles {
-		if _, pending := b.pendingCacheRefresh.Load(f); pending {
+		val, pending := b.pendingCacheRefresh.Load(f)
+		if !pending {
+			continue
+		}
+		ch := val.(chan struct{})
+		select {
+		case <-ch:
+		case <-ctx.Done():
 			return nil, false, nil
 		}
 	}
@@ -202,6 +213,8 @@ func (b *builtinServiceClient) incidentsFromCache(key incidentCacheKey, scopedFi
 
 // cacheRefreshWorker consumes from cacheRefreshChan and re-runs all parsed
 // conditions against the updated file content, storing results in the cache.
+// Each refresh gets a per-file child context so that it can be cancelled
+// when a newer change arrives for the same file or on shutdown.
 func (b *builtinServiceClient) cacheRefreshWorker() {
 	for {
 		select {
@@ -212,10 +225,23 @@ func (b *builtinServiceClient) cacheRefreshWorker() {
 			if !b.prepared {
 				continue
 			}
-			b.refreshCacheForFile(context.Background(), req.originalPath, req.contentPath)
-		case <-b.cacheRefreshDone:
+			fileCtx, fileCancel := context.WithCancel(b.cacheRefreshCtx)
+			b.activeRefreshes.Store(req.originalPath, fileCancel)
+			b.refreshCacheForFile(fileCtx, req.originalPath, req.contentPath)
+			b.activeRefreshes.Delete(req.originalPath)
+			fileCancel()
+		case <-b.cacheRefreshCtx.Done():
 			return
 		}
+	}
+}
+
+// completePendingRefresh removes the pending entry for a file and closes
+// its done channel, unblocking any Evaluate calls waiting in tryIncidentCache.
+func (b *builtinServiceClient) completePendingRefresh(filePath string) {
+	val, loaded := b.pendingCacheRefresh.LoadAndDelete(filePath)
+	if loaded {
+		close(val.(chan struct{}))
 	}
 }
 
@@ -288,7 +314,11 @@ func (b *builtinServiceClient) refreshCacheForFile(ctx context.Context, original
 	}
 
 	if !needsContent && !needsXML && !needsJSON {
-		b.pendingCacheRefresh.Delete(originalPath)
+		b.completePendingRefresh(originalPath)
+		return
+	}
+
+	if ctx.Err() != nil {
 		return
 	}
 
@@ -324,7 +354,7 @@ func (b *builtinServiceClient) refreshCacheForFile(ctx context.Context, original
 		}
 	}
 
-	b.pendingCacheRefresh.Delete(originalPath)
+	b.completePendingRefresh(originalPath)
 	b.log.V(5).Info("cache refreshed for working copy", "original", originalPath, "content", contentPath)
 }
 
@@ -487,7 +517,6 @@ func (b *builtinServiceClient) processFilecontentForFile(filePath string, conten
 			continue
 		}
 
-		// Check file pattern filter
 		if pc.condition.Filecontent.FilePattern != "" {
 			if !matchesFilePattern(pc.compiledFilePattern, pc.condition.Filecontent.FilePattern, relPath, baseName) {
 				continue
--- a/provider/internal/builtin/incident_cache.go
+++ b/provider/internal/builtin/incident_cache.go
@@ -196,8 +196,8 @@ func (b *builtinServiceClient) mergeIntoCache(key incidentCacheKey, filePath str
 // only include files in the provided scope set. Returns nil and false on cache miss.
 func (b *builtinServiceClient) incidentsFromCache(key incidentCacheKey, scopedFiles []string) ([]provider.IncidentContext, bool) {
 	b.incidentCacheMutex.RLock()
+	defer b.incidentCacheMutex.RUnlock()
 	fileMap, exists := b.incidentCache[key]
-	b.incidentCacheMutex.RUnlock()
 	if !exists {
 		return nil, false
 	}
@@ -260,6 +260,8 @@ func (b *builtinServiceClient) invalidateCacheForFile(filePath string) {
 // and updates the incident cache. The filePath should be the original file path
 // (not the working copy temp path) so cache keys align with Evaluate lookups.
 func (b *builtinServiceClient) refreshCacheForFile(ctx context.Context, originalPath string, contentPath string) {
+	defer b.completePendingRefresh(originalPath)
+
 	if !b.prepared || len(b.allParsedConditions) == 0 {
 		return
 	}
@@ -314,7 +316,6 @@ func (b *builtinServiceClient) refreshCacheForFile(ctx context.Context, original
 	}
 
 	if !needsContent && !needsXML && !needsJSON {
-		b.completePendingRefresh(originalPath)
 		return
 	}
 
@@ -354,7 +355,6 @@ func (b *builtinServiceClient) refreshCacheForFile(ctx context.Context, original
 		}
 	}
 
-	b.completePendingRefresh(originalPath)
 	b.log.V(5).Info("cache refreshed for working copy", "original", originalPath, "content", contentPath)
 }
 
--- a/provider/internal/builtin/incident_cache.go
+++ b/provider/internal/builtin/incident_cache.go
@@ -1,0 +1,566 @@
+package builtin
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/antchfx/jsonquery"
+	"github.com/antchfx/xmlquery"
+	"github.com/antchfx/xpath"
+	"github.com/konveyor/analyzer-lsp/provider"
+	"github.com/konveyor/analyzer-lsp/tracing"
+)
+
+// maxCacheFileSize is the maximum file size (in bytes) that will be read
+// during Prepare()'s incident caching. Files larger than this are skipped.
+const maxCacheFileSize = 100 * 1024 * 1024 // 100MB
+
+// cacheRefreshRequest is sent from the working copy manager to the
+// builtinServiceClient after a file change has been written to disk.
+type cacheRefreshRequest struct {
+	// originalPath is the real file path (used as cache key)
+	originalPath string
+	// contentPath is where the current content lives:
+	// - temp file for unsaved changes
+	// - original path for saved changes (re-read from disk)
+	contentPath string
+}
+
+// incidentCacheKey uniquely identifies a (capability, search params) pair.
+type incidentCacheKey struct {
+	cap    string
+	params string
+}
+
+// parsedCondition holds a pre-parsed, pre-compiled condition ready for execution
+// during Prepare()'s per-file processing. Compiling regexes and XPath expressions
+// once here avoids re-compilation for every file.
+type parsedCondition struct {
+	cacheKey            incidentCacheKey
+	cap                 string
+	condition           builtinCondition
+	compiledXPath       *xpath.Expr    // xml, xmlPublicID
+	compiledRegex       *regexp.Regexp // xmlPublicID
+	compiledFilePattern *regexp.Regexp // filecontent FilePattern, nil if empty or invalid regex
+	jsonXPath           string         // json
+}
+
+// makeIncidentCacheKey builds a canonical cache key for a condition.
+// Keys are derived from the parsed condition fields (Pattern, XPath, etc.),
+// not from raw conditionInfo bytes. This ensures Prepare() and Evaluate()
+// produce identical keys as long as YAML unmarshalling is deterministic,
+// which it is for these simple scalar/map fields.
+func makeIncidentCacheKey(cap string, cond builtinCondition) incidentCacheKey {
+	switch cap {
+	case "filecontent":
+		return incidentCacheKey{
+			cap:    cap,
+			params: fmt.Sprintf("%v-%v", cond.Filecontent.Pattern, cond.Filecontent.FilePattern),
+		}
+	case "xml":
+		return incidentCacheKey{
+			cap:    cap,
+			params: fmt.Sprintf("%v-%v", cond.XML.XPath, sortedNamespaces(cond.XML.Namespaces)),
+		}
+	case "xmlPublicID":
+		return incidentCacheKey{
+			cap:    cap,
+			params: fmt.Sprintf("%v-%v", cond.XMLPublicID.Regex, sortedNamespaces(cond.XMLPublicID.Namespaces)),
+		}
+	case "json":
+		return incidentCacheKey{
+			cap:    cap,
+			params: cond.JSON.XPath,
+		}
+	default:
+		return incidentCacheKey{cap: cap}
+	}
+}
+
+func sortedNamespaces(ns map[string]string) string {
+	if len(ns) == 0 {
+		return ""
+	}
+	keys := make([]string, 0, len(ns))
+	for k := range ns {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	var b strings.Builder
+	for i, k := range keys {
+		if i > 0 {
+			b.WriteByte('|')
+		}
+		b.WriteString(k)
+		b.WriteByte('=')
+		b.WriteString(ns[k])
+	}
+	return b.String()
+}
+
+// isChainedCondition returns true if the condition has non-empty Filepaths,
+// which are template-expanded at Evaluate() time and cannot be pre-cached.
+func isChainedCondition(cap string, cond builtinCondition) bool {
+	switch cap {
+	case "filecontent":
+		return len(cond.Filecontent.Filepaths) > 0
+	case "xml":
+		return len(cond.XML.Filepaths) > 0
+	case "xmlPublicID":
+		return len(cond.XMLPublicID.Filepaths) > 0
+	case "json":
+		return len(cond.JSON.Filepaths) > 0
+	default:
+		return false
+	}
+}
+
+// tryIncidentCache attempts to look up cached incidents for a capability.
+// Returns the incidents, whether a cache hit occurred, and any error.
+func (b *builtinServiceClient) tryIncidentCache(cap string, cond builtinCondition, patterns []string) ([]provider.IncidentContext, bool, error) {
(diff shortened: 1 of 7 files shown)


provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+// Keys are derived from the parsed condition fields (Pattern, XPath, etc.),
+// not from raw conditionInfo bytes. This ensures Prepare() and Evaluate()
+// produce identical keys as long as YAML unmarshalling is deterministic,
+// which it is for these simple scalar/map fields.
+func makeIncidentCacheKey(cap string, cond builtinCondition) incidentCacheKey {
+	switch cap {
+	case "filecontent":
+		return incidentCacheKey{
+			cap:    cap,
+			params: cond.Filecontent.Pattern + "\x00" + cond.Filecontent.FilePattern,
```
shawn-hurley: nit: fmt.Sprintf("%v-%v") or something similar rather than null byte and doing string concatanation

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+	// Single filesystem walk
+	files, err := searcher.Search(provider.SearchCriteria{})
+	if err != nil {
+		b.log.Error(err, "failed to build file index in Prepare, falling back to per-Evaluate walks")
+		return nil
+	}
+
+	b.fileIndex = files
+	b.allParsedConditions = parsedConds
+	b.incidentCache = make(map[incidentCacheKey]map[string][]provider.IncidentContext)
```
shawn-hurley, line 521: If you have a mutex guarding this, then we should lock and unlock here, I think?

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+
+	// Process all files through worker pool to populate the incident cache.
+	if len(parsedConds) > 0 && len(files) > 0 {
+		const numWorkers = 5
```
shawn-hurley: I wonder if we can make this configurable with a providerSpecificConfig? 

For now, maybe just add a todo in the code?

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+		fileChan := make(chan string, 10)
+		var wg sync.WaitGroup
```
shawn-hurley: this gets the zero value by default, so it should work, I would just prefer being explict with `wg := sync.WaitGroup{}`

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+	defer b.incidentCacheMutex.Unlock()
+	if b.incidentCache[key] == nil {
+		b.incidentCache[key] = make(map[string][]provider.IncidentContext)
+	}
+	b.incidentCache[key][filePath] = append(b.incidentCache[key][filePath], incidents...)
+}
+
+// incidentsFromCache looks up cached incidents for a cache key, filtered to
+// only include files in the provided scope set. Returns nil and false on cache miss.
+func (b *builtinServiceClient) incidentsFromCache(key incidentCacheKey, scopedFiles []string) ([]provider.IncidentContext, bool) {
```
shawn-hurley: Starting to wonder if all the cache-specific functions should be in a different file?

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+		}
+	}
+	return result, true
+}
+
+// processFileForAllCaps reads a file once and runs all applicable pre-compiled
+// condition searches against it, merging results into the incident cache.
+func (b *builtinServiceClient) processFileForAllCaps(ctx context.Context, filePath string) {
+	// Skip files that are too large
+	const maxFileSize = 100 * 1024 * 1024 // 100MB
```
shawn-hurley: Let's move the const to the top of the file?

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+			b.log.V(5).Error(err, "failed to parse JSON during Prepare", "file", filePath)
+		} else {
+			b.processJSONForFile(ctx, filePath, doc)
+		}
+	}
+}
+
+// processFilecontentForFile runs all filecontent conditions against pre-read content.
+func (b *builtinServiceClient) processFilecontentForFile(filePath string, content []byte) {
+	relPath, err := filepath.Rel(b.config.Location, filePath)
```
shawn-hurley: We already did this, so I don't know if we need to do it again, unless I am missing something?

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+	if err != nil {
+		b.log.V(5).Error(err, "failed to stat file during Prepare", "file", filePath)
+		return
+	}
+	if fileInfo.Size() > maxFileSize {
+		b.log.V(5).Info("skipping large file during Prepare", "file", filePath, "size", fileInfo.Size())
+		return
+	}
+
+	// Compute relative path from base location for pattern matching
```
shawn-hurley: We probably should do this work before and only do the fileInfo bits if/when we need to read a file. 

This is because IIRC, the os.Stat has to make a system call, so if we know we don't have any conditions that require searching that file, we don't need to make that system call.

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+			needsContent = true
+		case "xml", "xmlPublicID":
+			if isXML {
+				needsXML = true
+			}
+		case "json":
+			if isJSON {
+				needsJSON = true
+			}
+		}
```
shawn-hurley: Could we do the logic, or call the functions, in this loop?

I think that in the case that there is no match, it will continue just as fast, and then we don't have to re-loop all the parsed conditions?

provider/internal/builtin/service_client.go
```diff
@@ -103,26 +913,35 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
...
-		},
-		RuleScopeConstraints: provider.IncludeExcludeConstraints{
-			IncludePathsOrPatterns: includedPaths,
-			ExcludePathsOrPatterns: excludedPaths,
-		},
-		FailFast: true,
-		Log:      p.log,
+	// searchFiles resolves file lists from the cached index when available,
+	// otherwise falls back to a full filesystem walk.
+	searchFiles := func(criteria provider.SearchCriteria) ([]string, error) {
```
shawn-hurley, line 672: We have to handle working copies, I think, in this flow, and for that file, we do have to re-run and update the conditions that now apply or don't apply, I think.

provider/internal/builtin/service_client.go
```diff
@@ -160,11 +979,28 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
 			return response, fmt.Errorf("could not parse provided regex pattern as string: %v", conditionInfo)
 		}
 
+		// Try incident cache lookup
+		if incidents, ok, err := p.tryIncidentCache(cap, cond, func() []string {
```
shawn-hurley: lets just do this logic and save it in a variable rather then call an inline function to pass as a parameter?

provider/internal/builtin/service_client.go
```diff
@@ -177,32 +1013,27 @@ func (p *builtinServiceClient) Evaluate(ctx context.Context, cap string, conditi
...
 		return response, nil
 	case "xml":
 		query, err := xpath.CompileWithNS(cond.XML.XPath, cond.XML.Namespaces)
 		if query == nil || err != nil {
 			return response, fmt.Errorf("could not parse provided xpath query '%s': %v", cond.XML.XPath, err)
 		}
-		xmlFiles, err := fileSearcher.Search(provider.SearchCriteria{
+
+		// Try incident cache lookup
+		if incidents, ok, err := p.tryIncidentCache(cap, cond, []string{"*.xml", "*.xhtml"}); err != nil {
```
shawn-hurley: I don't know if this is exactly the right logic for dealing with working copy mgr like above.

provider/internal/builtin/service_client.go
shawn-hurley, line 634: This is where the working copy manager updates the file content. I think we also need to invalidate the file in the cache and use the updated file as the cache search content. 

Let me know if that doesn't make sense :)

shawn-hurley (changes requested):
This is really good so far, some minor style nits, but:

1. Thank you for breaking stuff out into functions. I think that may make it easier to break this giant file into more isolated files, just for readability.
2. The cache implementation is really good; there is just one gotcha that I think we have to cover.

The comment about the working copy manager, I think, is that gotcha. 

Basically, the working copy manager can take and store "updated" content for a file, even if it isn't saved on disk. We then attempt to use that updated content when searching. 

What this means for the cache is that when we receive a notification of a file change, we need to pass the updated content and filename to one of the workers (which must stay running) to update the cache. 

Once the cache is updated, I think we can just use it. The reformat incidents for the working copy should happen when we set the cache.

Let me know if this makes sense.

Note: this behavior is new to me, until a couple of weeks ago, so I may be missing something here. If we need to get on a quick call, we can set up a time.

provider/internal/builtin/service_client.go
savitharaghunathan: added a todo. will create an issue for this

provider/internal/builtin/service_client.go
savitharaghunathan: moved these to a different file

provider/internal/builtin/service_client.go
```diff
@@ -43,43 +44,852 @@ type builtinServiceClient struct {
...
+	key := makeIncidentCacheKey(cap, cond)
+	scopedFiles, err := b.filterCachedFiles(cond, provider.SearchCriteria{
+		Patterns: patterns,
+	})
+	if err != nil {
+		return nil, false, err
+	}
+	incidents, ok := b.incidentsFromCache(key, scopedFiles)
+	return incidents, ok, nil
+}
```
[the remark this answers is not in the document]
savitharaghunathan: @coderabbitai review

provider/internal/builtin/service_client.go
savitharaghunathan: updated

provider/internal/builtin/service_client.go
savitharaghunathan: updated

provider/internal/builtin/service_client.go
savitharaghunathan: using `maxCacheFileSize` and moved it to the incident_cache.go file

provider/internal/builtin/service_client.go
savitharaghunathan: updated

provider/internal/builtin/service_client.go
savitharaghunathan: updated

provider/internal/builtin/service_client.go
savitharaghunathan: upadted

provider/internal/builtin/service_client_test.go
```diff
@@ -1368,3 +1370,858 @@ func Test_builtinServiceClient_performFileContentSearch_Multiline(t *testing.T)
...
+			}
+
+			resultWithPrepare, errWithPrepare := clientWithPrepare.Evaluate(
+				context.TODO(), tt.capability, condBytes)
+
+			// Compare errors
+			if (errNoPrepare != nil) != (errWithPrepare != nil) {
+				t.Fatalf("error mismatch: noPrepare=%v, withPrepare=%v",
+					errNoPrepare, errWithPrepare)
+			}
```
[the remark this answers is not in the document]
savitharaghunathan: fixed. See commit 048b556.

provider/internal/builtin/service_client.go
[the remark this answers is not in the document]
savitharaghunathan: fixed. see commit 90a536c.

provider/internal/builtin/service_client.go
savitharaghunathan: updated

provider/internal/builtin/incident_cache.go
```diff
@@ -0,0 +1,579 @@
...
+	key := makeIncidentCacheKey(cap, cond)
+	scopedFiles, err := b.filterCachedFiles(cond, provider.SearchCriteria{
+		Patterns: patterns,
+	})
+	if err != nil {
+		return nil, false, err
+	}
+	// If any scoped file is pending cache refresh, fall back to
+	// filesystem walk so Evaluate doesn't return stale/incomplete results.
+	for _, f := range scopedFiles {
```
shawn-hurley, line 143: Could this wait until the cache is re-populated with the results?

provider/internal/builtin/incident_cache.go
```diff
@@ -0,0 +1,579 @@
...
+		if incidents, ok := fileMap[f]; ok {
+			result = append(result, incidents...)
+		}
+	}
+	return result, true
+}
+
+// cacheRefreshWorker consumes from cacheRefreshChan and re-runs all parsed
+// conditions against the updated file content, storing results in the cache.
+func (b *builtinServiceClient) cacheRefreshWorker() {
```
shawn-hurley, line 218: nit: can we add a context so that we can also cancel these if needed?

Say a request for Notify changes comes in for file A, we start working, and before we are done, a new one for file A comes in, I think we should cancel these workers and start over. Just an example, but also when we shutdown I think it would be good for the context to be able to stop the go routine.

provider/internal/builtin/service_client.go
```diff
@@ -89,6 +568,16 @@ func (p *builtinServiceClient) NotifyFileChanges(ctx context.Context, changes ..
 			filtered = append(filtered, change)
 		}
 	}
+	// Invalidate cache and mark files as pending refresh so that
+	// Evaluate falls back to filesystem walk until the refresh completes.
+	if p.prepared {
+		for _, change := range filtered {
+			p.invalidateCacheForFile(change.Path)
+			p.pendingCacheRefresh.Store(change.Path, true)
```
shawn-hurley: nit: reverse the order here, mark as pending then invalidate. 

The worry is a timing issue where a request comes in between the two and returns zero results.

provider/internal/builtin/wcmanger.go
```diff
@@ -116,18 +120,30 @@ func (t *workingCopyManager) startWorker() {
...
-			if change.Saved && wcExists {
-				t.wcMutex.Lock()
-				delete(t.workingCopies, change.Path)
-				t.wcMutex.Unlock()
-				if _, err := os.Stat(change.Path); err == nil || !os.IsNotExist(err) {
-					err := os.Remove(wcPath)
-					if err != nil {
-						t.log.Error(err, "failed to remove working copy")
+			if change.Saved {
+				if wcExists {
```
shawn-hurley: Keep this if clause in the block above?

provider/internal/builtin/service_client.go
```diff
@@ -89,6 +568,16 @@ func (p *builtinServiceClient) NotifyFileChanges(ctx context.Context, changes ..
 			filtered = append(filtered, change)
 		}
 	}
+	// Invalidate cache and mark files as pending refresh so that
+	// Evaluate falls back to filesystem walk until the refresh completes.
+	if p.prepared {
```
shawn-hurley, line 603: I don't know, but I think you may need to start the refresh workers here?

shawn-hurley (changes requested):
This is getting really close, I think, getting the tests to pass and cleaning up some of the nits, and we should be good to go! This is excellent, thank you!

provider/internal/builtin/wcmanger.go
savitharaghunathan: @shawn-hurley can you explain what is needed here? do you mean I keep it as `if change.Saved && wcExists {` instead of the nested ifs?

shawn-hurley: Yeah instead of nesting the if's

savitharaghunathan: updated

provider/internal/builtin/service_client.go
```diff
@@ -51,35 +89,506 @@ type fileTemplateContext struct {
...
+				for file := range fileChan {
+					b.processFileForAllCaps(ctx, file)
+				}
+			})
+		}
+
+		for _, file := range files {
+			fileChan <- file
+		}
+		close(fileChan)
```
shawn-hurley, line 539: Will this close stop the execution of the workers?

savitharaghunathan: my understanding is that it wont stop the execution abruptly but will not take any new items. Please correct me if I am wrong in my thoughts.

savitharaghunathan: do I need to explicitly close the workers here?

savitharaghunathan: stop execution*

provider/internal/builtin/incident_cache.go
```diff
@@ -0,0 +1,608 @@
...
+
+// processFilecontentForFile runs all filecontent conditions against pre-read content.
+func (b *builtinServiceClient) processFilecontentForFile(filePath string, content []byte, relPath string, baseName string) {
+	for i := range b.allParsedConditions {
+		pc := &b.allParsedConditions[i]
+		if pc.cap != "filecontent" {
+			continue
+		}
+
+		if pc.condition.Filecontent.FilePattern != "" {
```
shawn-hurley, line 417: I might suggest that these inner bits of actaully running a condition get their own function and can be shared with the processFileForAllCaps code?

minor nit shouldn't block this.

provider/internal/builtin/service_client.go
```diff
@@ -51,35 +89,506 @@ type fileTemplateContext struct {
...
+	if err != nil {
+		b.log.Error(err, "failed to build file index in Prepare, falling back to per-Evaluate walks")
+		return nil
+	}
+
+	b.fileIndex = files
+	b.allParsedConditions = parsedConds
+	b.incidentCacheMutex.Lock()
+	b.incidentCache = make(map[incidentCacheKey]map[string][]provider.IncidentContext)
+	b.incidentCacheMutex.Unlock()
```
shawn-hurley: should this be defered?

provider/internal/builtin/service_client.go
```diff
@@ -51,35 +89,506 @@ type fileTemplateContext struct {
...
+	b.prepared = true
+	b.log.V(5).Info("Prepare complete", "fileCount", len(files), "parsedConditions", len(parsedConds))
+	return nil
+}
+
+// startCacheRefreshWorkers initialises the cache refresh channel and spawns
+// worker goroutines the first time it is called. Subsequent calls are no-ops.
+// Safe to call concurrently from Prepare and NotifyFileChanges.
+func (b *builtinServiceClient) startCacheRefreshWorkers() {
+	b.cacheRefreshMu.Lock()
```
shawn-hurley: We don't need to lock and unlock here, this would isn't mutating the map

provider/internal/builtin/service_client.go
```diff
@@ -51,35 +89,506 @@ type fileTemplateContext struct {
...
+	b.workingCopyMgr.cacheRefreshChan = b.cacheRefreshChan
+	for range numWorkers {
+		go b.cacheRefreshWorker()
+	}
+}
+
+// stopCacheRefreshWorkers cancels running workers and resets state so
+// startCacheRefreshWorkers can re-initialise on the next call.
+func (b *builtinServiceClient) stopCacheRefreshWorkers() {
+	b.cacheRefreshMu.Lock()
```
shawn-hurley: Same here, no need to lock/unlock

provider/internal/builtin/service_client.go
```diff
@@ -51,35 +89,506 @@ type fileTemplateContext struct {
...
+		b.cacheRefreshCancel()
+		b.cacheRefreshCancel = nil
+		b.cacheRefreshChan = nil
+		b.workingCopyMgr.cacheRefreshChan = nil
+	}
+}
+
+func (b *builtinServiceClient) Stop() {
+	b.workingCopyMgr.stop()
+	b.stopCacheRefreshWorkers()
```
shawn-hurley: nit: I would just call cancel and nil things here.

shawn-hurley (changes requested):
Just the extra locks and maybe a defer unlock that isn't but otherwise I think this looks good!

provider/internal/builtin/service_client.go
savitharaghunathan: Done — added `incidentCacheMutex.Lock()/Unlock()` around the `make()` call (service_client.go:525-527). Also guarded in the reset block at lines 409-411.

provider/internal/builtin/service_client.go
savitharaghunathan: Good call — restructured `processFileForAllCaps` with a lazy `loadContent` closure (incident_cache.go:376). `os.Stat` and file read only happen when a condition actually needs the file content (e.g., a filecontent FilePattern matches, or extension is .xml/.json).

provider/internal/builtin/service_client.go
savitharaghunathan: Updated — conditions are now parsed once in `Prepare()` and stored as `allParsedConditions`. `processFileForAllCaps` (incident_cache.go:404-509) processes all conditions per file in a single loop with inline execution — no second pass needed.

provider/internal/builtin/service_client.go
savitharaghunathan: Working copies are handled through the cache refresh mechanism:

- `NotifyFileChanges` invalidates the cache for the changed file and marks it pending (service_client.go:607-633)
- For unsaved changes, WCM writes a temp file and sends a `cacheRefreshRequest` (wcmanger.go:154-162)
- `refreshCacheForFile` re-runs all conditions against the updated content, storing results under the original path (incident_cache.go:262-359)
- `tryIncidentCache` waits for pending refreshes before returning results (incident_cache.go:143-154)

For the `file` cap specifically, `searchFiles` uses `filterCachedFiles` which returns original paths from the index — since the `file` cap only matches on filenames (not content), this gives the correct result regardless of working copy state.

provider/internal/builtin/service_client.go
savitharaghunathan: Done — changed `tryIncidentCache` to take `patterns []string` directly instead of a closure (incident_cache.go:130). Callers now build the patterns slice as a local variable before passing it in.

provider/internal/builtin/service_client.go
savitharaghunathan: Working copy handling for XML (and all content-aware caps) goes through the cache refresh path — `refreshCacheForFile` reads from the WC temp file, re-parses the XML DOM, and re-runs all applicable conditions (incident_cache.go:340-347). The original path is used as the cache key so `tryIncidentCache` lookups align correctly.

provider/internal/builtin/service_client.go
savitharaghunathan: Done — `NotifyFileChanges` now:
1. Invalidates all cache entries for the file (`invalidateCacheForFile`, incident_cache.go:250-257)
2. Marks the file as pending refresh (`pendingCacheRefresh`)
3. For saved changes: sends a refresh request directly (service_client.go:626-631)
4. For unsaved changes: WCM writes the temp file and sends a refresh request (wcmanger.go:154-162)
5. `refreshCacheForFile` re-runs all conditions against the updated content and stores results under the original path

provider/internal/builtin/service_client.go
savitharaghunathan: Done — `Stop()` now inlines the cancel and nil logic directly (service_client.go:582-590).

provider/internal/builtin/service_client.go
savitharaghunathan: removed it as deffering would cause a deadlock when `mergeIntoCache` is called

provider/internal/builtin/incident_cache.go
savitharaghunathan: done

shawn-hurley (approved):

shawn-hurley:
I had to rebase, because of the changes with the refactoring of the providers I believe. 

Hopefully this allows CI to pass

provider/internal/builtin/incident_cache.go
```diff
@@ -0,0 +1,505 @@
...
+				return
+			}
+			if !b.prepared {
+				continue
+			}
+			fileCtx, fileCancel := context.WithCancel(b.cacheRefreshCtx)
+			b.activeRefreshes.Store(req.originalPath, fileCancel)
+			b.refreshCacheForFile(fileCtx, req.originalPath, req.contentPath)
+			b.activeRefreshes.Delete(req.originalPath)
+			fileCancel()
```
pranavgaikwad, line 232: are you potenitally calling cancel func twice? once in Store() and here?

provider/internal/builtin/service_client.go
```diff
@@ -43,6 +43,44 @@ type builtinServiceClient struct {
...
+	incidentCacheMutex sync.RWMutex
+	// allParsedConditions holds pre-compiled conditions from Prepare(),
+	// deduplicated by cache key. Chained conditions are excluded.
+	allParsedConditions []parsedCondition
+
+	// cacheRefreshChan receives requests from the working copy manager
+	// after a file change has been written to disk. Workers on the
+	// builtinServiceClient side consume from this channel and re-run
+	// conditions against the updated content.
+	cacheRefreshChan chan cacheRefreshRequest
```
pranavgaikwad, line 64: what happens when changes are not written to the disk? do we get the correct results for a given in-memory file?

savitharaghunathan: I think so. Unsaved changes are handled via the working copy manager, it writes the in-memory content to a temp file, then sends a cache refresh request. The refresh worker re-runs all conditions against that temp file and stores results keyed by the original path. Evaluate blocks on the pending channel until the refresh completes..

provider/internal/builtin/incident_cache.go
savitharaghunathan: Ya... I wasnt sure if I could defer bec of the loop and how the call would stack between contexts

pranavgaikwad (approved):

PR #1132, opened 2026-03-23
:bug: Quote filepaths when passing them to yaml processing during cha…

jmle:
…ining (#1122)

When passing around filepaths for condition chaining, if any of the files has special characters (for instance, `]`), the generated yaml is not valid. We need to quote the filepaths to prevent this from happening.

Fixes https://github.com/konveyor/analyzer-lsp/issues/1082

* **Bug Fixes**
* Enhanced serialization of file paths in conditions to properly handle special characters, spaces, and quotes.

* **Tests**
* Added comprehensive test coverage for edge cases involving special characters in file path formatting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 target/
 *.metadata*
 jdt.ls-java-project
+examples/c-sharp-dummy-example/obj/
 examples/java/*.project
 examples/python/.venv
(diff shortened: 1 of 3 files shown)


mguetta1 (approved):
LGTM

PR #1133, opened 2026-03-25
:bug: Fixes Respect Keyboard interrupt kantra#776

jwmatthews:
See: https://github.com/konveyor/kantra/issues/776

  analyzer-lsp fixes:
  - Buffer engine response channel so workers don't block after handler exits
  - Guard rule dispatch loop with ctx.Done() to prevent deadlock on full channel
  - Buffer ProviderStart abConfigChan and add ctx.Done() to init wait select
  - Add 5s timeout to gRPC provider Stop() (was context.TODO() with no timeout)
  - Thread context through GetProviderClient/NewGRPCClient for signal-aware startup
  - Rewrite checkServicesRunning to accept context and fix broken 30s timeout

Related to https://github.com/konveyor/kantra/pull/777

Co-Authored-By: Claude Opus 4.6 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/core/analyzer.go
+++ b/core/analyzer.go
@@ -141,7 +141,7 @@ func (a *analyzer) ProviderStart() error {
 		Message: "Starting provider init",
 		Total:   len(a.providers),
 	})
-	abConfigChan := make(chan []provider.InitConfig)
+	abConfigChan := make(chan []provider.InitConfig, len(a.providers))
 	providerInitCtx, cancelFunc := context.WithCancel(a.ctx)
 	waitGroup := sync.WaitGroup{}
 	go func() {
@@ -202,12 +202,20 @@ func (a *analyzer) ProviderStart() error {
 	}
 
 	if noTimeout {
-		<-c
-		a.log.V(3).Info("started all non builtin providers")
+		select {
+		case <-c:
+			a.log.V(3).Info("started all non builtin providers")
+		case <-a.ctx.Done():
+			cancelFunc()
+			return fmt.Errorf("provider init cancelled: %w", a.ctx.Err())
+		}
 	} else {
 		select {
 		case <-c:
 			a.log.V(3).Info("started all non builtin providers")
+		case <-a.ctx.Done():
+			cancelFunc()
+			return fmt.Errorf("provider init cancelled: %w", a.ctx.Err())
 		case <-time.After(timeout):
 			cancelFunc()
 			return fmt.Errorf("timed out starting providers after %s", timeout)
--- a/core/analyzer.go
+++ b/core/analyzer.go
@@ -141,14 +141,13 @@ func (a *analyzer) ProviderStart() error {
 		Message: "Starting provider init",
 		Total:   len(a.providers),
 	})
-	abConfigChan := make(chan []provider.InitConfig, len(a.providers))
+	abConfigChan := make(chan []provider.InitConfig)
 	providerInitCtx, cancelFunc := context.WithCancel(a.ctx)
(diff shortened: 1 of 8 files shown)


core/analyzer.go
```diff
@@ -202,12 +205,20 @@ func (a *analyzer) ProviderStart() error {
 	}
 
 	if noTimeout {
-		<-c
-		a.log.V(3).Info("started all non builtin providers")
+		select {
+		case <-c:
+			a.log.V(3).Info("started all non builtin providers")
+		case <-a.ctx.Done():
```
shawn-hurley, line 211: We should add timeout to the providerInitCtx, and then update this to use the providerInitCtx. 

It should be built off to a.ctx so that if a.ctx is done, it will provide providerInitCtx, and providerInitCtx timeout is a part of the context. Then we can remove the if/else and just have a single select on c or providerInitCtx.Done().

Let me know if that makese sense or not?

core/analyzer.go
```diff
@@ -179,7 +178,11 @@ func (a *analyzer) ProviderStart() error {
 					Current: i + 1,
 					Total:   len(a.providers),
 				})
-				abConfigChan <- additionalBuiltins
+				select {
```
shawn-hurley, line 181: This is fixing a channel drain on when context is done. 


But the fix I don't think is achieving the same result, we are now marking the wait group finished when additional builtins are passed to the channel, not when the the go routine that adds the additonal builtins to the slice is completed. 

If the channel has a buffer of 1. there will still be a lag between updating the slice and the WaitGroup.Wait() finishing, and this I think could cause a race condition.

the waitGroup.Done() I think needs to be called original location, we just need to add on cancel to all call waitGroup.Done(). 

For draining the waitGroup, there could be at most 1 other object in the buffer. 

But, I think that the fix below, around adding context to if noTimeout, actually fully solves the issue that you are trying to solve, without having to worry about this.

engine/engine.go
```diff
@@ -360,8 +360,14 @@ func (r *ruleEngine) RunRulesScopedWithOptions(ctx context.Context, ruleSets []R
 		rule.conditionContext = newContext
 		rule.scope = scopes
 		rule.carrier = carrier
-		r.ruleProcessing <- rule
+		select {
```
shawn-hurley, line 363: a couple of things:

1. I think I would be hard-pressed to agree on using the goTo here. 
2. Why add a buffer to the return channel? 


To add support for canceling w/ the working group, we should follow a similar pattern in analyzer.go where we have the wait in a go routine, and when we finish, we close a sentinel channel, with a select that selects on ctx.Done() or the close of the channel.

provider/lib/lib.go
```diff
@@ -9,11 +11,11 @@ import (
 )
 
 // We need some wrapper that can deal with out of tree providers, this will be a call, that will mock it out, but go against in tree.
-func GetProviderClient(config provider.Config, log logr.Logger, progress *progress.Progress) (provider.InternalProviderClient, error) {
+func GetProviderClient(ctx context.Context, config provider.Config, log logr.Logger, progress *progress.Progress) (provider.InternalProviderClient, error) {
```
shawn-hurley, line 14: I don't know if we need to do this or if we want to do this,

But this will break consumers of the lib package.

Can we add a `GetProviderClientWithContext` function and update this function to call it with the context set to background? 

What are your thoughts?

shawn-hurley (changes requested):
This is really good and helpful. This has been a nagging issue that I think should be solved. 

Thank you!

PR #1142, opened 2026-03-31
:sparkles: Refactor generic provider into individual providers

### LINKED ISSUE #1013: [refactor]: Generic Service Client to have specific providers.
We are not finding it the case, that we can have a generic client that reliably interfaces with any generic language server. The initialization of those language servers are to different to make this fully compatible. Because of this, we should consider breaking the providers that are in the generic provider out to their specific languages.

jmle:
- Implements [this enhancement](https://github.com/konveyor/enhancements/tree/master/enhancements/generic-to-specific-providers)
- Provided implementation plan: each commit is roughly a step of the plan
- Fixes https://github.com/konveyor/analyzer-lsp/issues/1013

## Summary by CodeRabbit

* **New Features**
  * Added language-specific external providers for Go, Python, and Node.js (replacing the prior generic provider).

* **Documentation**
  * Updated guides, setup, testing, and provider configuration docs with per-language examples, e2e test paths, and a migration plan.

* **Chores**
  * Updated build/test targets, demo/test run commands, container images, and CI workflows to use the new per-language providers and removed legacy generic-provider artifacts.


.github/workflows/image-build.yaml
```diff
@@ -49,6 +49,19 @@ jobs:
             containerfile: "./external-providers/yq-external-provider/Dockerfile"
             context: "."
             architectures: '[ "amd64", "arm64" ]'
+          - name: go-external-provider
+            containerfile: "./external-providers/go-external-provider/Dockerfile"
+            context: "."
+            architectures: '[ "amd64", "arm64" ]'
+            extra-args: "--build-arg GOLANG_DEP_IMAGE=quay.io/konveyor/golang-dependency-provider:${{ needs.compute-deps-refs.outputs.tag }}"
```
shawn-hurley: Could this just be built in the go provider and not be its own command?

external-providers/go-external-provider/pkg/go_external_provider/debug-go.log
```diff
@@ -0,0 +1,60 @@
+[Trace - 15:09:27.715 PM] Sending request 'initialize - (1)'.
```
shawn-hurley: proabably should remove this

external-providers/go-external-provider/pkg/go_external_provider/service_client.go
```diff
@@ -0,0 +1,148 @@
...
+	eval, err := base.NewLspServiceClientEvaluator[*GoServiceClient](sc, g.GetGoServiceClientCapabilities(log))
+	if err != nil {
+		return nil, fmt.Errorf("lsp service client evaluator error: %w", err)
+	}
+	sc.LSPServiceClientEvaluator = eval
+
+	return sc, nil
+}
+
+func (g *GoServiceClientBuilder) GetGoServiceClientCapabilities(log logr.Logger) []base.LSPServiceClientCapability {
```
shawn-hurley, line 97: This wouldn't be used right because in the provider, you are just handing them back?

external-providers/go-external-provider/pkg/go_external_provider/service_client.go
```diff
@@ -0,0 +1,148 @@
...
+	} else {
+		caps = append(caps, base.LSPServiceClientCapability{
+			Capability: echoCap,
+			Fn:         serviceClientFn((*GoServiceClient).EvaluateEcho),
+		})
+	}
+	return caps
+}
+
+type echoCondition struct {
```
shawn-hurley: Can we drop this as well?

shawn-hurley (approved):
High level, this looks right to me, and thank you for updating the documentation!

I have one bigger question: can/should we refactor the base?SLPServiceClient code to make it clearer what it is giving you? 

I don't know if it makes sense to do this now, as a follow-up or not, but I am just wondering if now that they are all not in a single place, it would be clear that the base client gives you a default Evaluate method, for instance, that does XXX? 

If we say the base class just has a generic `BaseEvaluate` method, would that make it clear for the given service client to call it?

I think it could also make the capabilities a bit clearer.

If folks disagree, I don't have that strong an opinion, but I wanted to start the conversation.

.github/workflows/image-build.yaml
jmle: I guess you mean removing the go dependency provider and baking it into the go provider right? Yeah totally, I was actually thinking the same

eemcmullan (approved):

PR #1191, opened 2026-09-21
:bug: Force Maven wagon transport when insecure repos are enabled

### LINKED ISSUE #1190: Maven "insecure" option is a no-op: -Dmaven.wagon.http.ssl.insecure is ignored by Maven 3.9 native transport
## Summary

The java-external-provider's Maven **"insecure"** option is intended to let analysis
proceed against Maven repositories whose TLS certificate is not trusted (self-signed or
signed by a private CA). It currently has **no effect**.

The option is implemented by passing `-Dmaven.wagon.http.ssl.insecure=true` to Maven, but
that property only affects the legacy **Wagon** HTTP transport. Since Maven **3.9.0** the
default resolver transport is **native** (not Wagon), so the property is silently ignored.
The provider bundle ships `apache-maven-3.9.15`, so the flag never disables TLS
verification and PKIX failures still abort dependency resolution.

## Affected code

The insecure toggle appends `-Dmaven.wagon.http.ssl.insecure=true` in the
java-external-provider:

- `bldtool/maven.go` — `dependency:tree`
- `dependency/maven_resolver.go` — `go-offline-maven-plugin:resolve-dependencies`
- `maven_downloader.go`

It is also **not propagated to the JDTLS / m2e side**. Only `userSettings` and
`globalSettings` are passed to the language server on init, so even if the flag worked for
the CLI dependency steps, the m2e project import inside JDTLS would still verify TLS.

## Why it is a no-op

- Maven 3.9.0 changed the default resolver transport from Wagon to native HTTP:
  - https://maven.apache.org/docs/3.9.0/release-notes.html
  - https://maven.apache.org/guides/mini/guide-resolver-transport.html
- `maven.wagon.http.ssl.insecure` is a **Wagon-only** property. Under the native transport
  it does nothing.
- The bundle installs `apache-maven-3.9.15`, which uses the native transport by default.

## Impact

When a user runs analysis "insecure" against a Maven repository whose TLS certificate is
not trusted by the JVM truststore, dependency resolution still fails with a PKIX
`unable to find valid certification path to requested target` error. Because the provider
logs `unable to resolve sources, continuing...` and proceeds anyway, the analysis
**reports success** while the Java rules find nothing (the JDTLS symbol index is empty) —
producing silently incomplete/empty results, which is the worst failure mode.

## Suggested fixes (any / all)

1. When insecure is requested, also force the Wagon transport so the existing property
   applies: pass `-Dmaven.resolver.transport=wagon` alongside
   `-Dmaven.wagon.http.ssl.insecure=true`. (Simplest — makes the existing flag actually
   work.)
2. Or configure the native transport's equivalent relaxed-TLS mode (Maven Resolver
   `aether.connector.https.securityMode=insecure` — needs verification for 3.9.x).
3. Propagate the insecure behavior to the JDTLS / m2e side as well, otherwise project
   import will still fail even after the CLI steps succeed.
4. Longer term, prefer proper CA trust over disabling verification (tracked separately as
   an RFE in `konveyor/tackle2-hub`).

## Related

- `konveyor/tackle2-addon-analyzer#92` — closed insecure-repo bug; it was noted there that
  the option conflated HTTP-blocking with TLS verification.

Koncur PR: 125

dymurray:
## Problem

The insecure Maven option (`mvn.insecure.enabled`) is meant to allow analysis against a Maven repository whose TLS certificate is not trusted (self-signed or signed by a private CA). It was implemented solely by passing `-Dmaven.wagon.http.ssl.insecure=true`, which only affects the legacy **Wagon** HTTP transport.

Since **Maven 3.9.0** the default resolver transport is the **native** HTTP transport, which ignores that property. As a result the flag was a no-op:

- dependency resolution still failed PKIX (`unable to find valid certification path`),
- sources were not resolved (`unable to resolve sources, continuing...`),
- `help:evaluate` failed (`could not get maven local repo path: exit status 1`),
- and analysis silently produced empty/incomplete Java results while reporting success.

Fixes #1190.

## Fix

Force the Wagon transport whenever insecure is requested, so the existing insecure flag actually takes effect. The two flags now live in a single shared slice (`dependency.MavenInsecureArgs`) and are applied in the two independent places that each reproduce the failure:

1. **The provider's own `mvn` subprocess invocations** — `dependency:tree`, go-offline resolve, `dependency:copy`, and `help:evaluate`. The last (`getMavenLocalRepoPath`) previously appended no insecure flag at all, which is the direct cause of the `could not get maven local repo path` error.

2. **The JDTLS JVM** — so its embedded m2e resolver (bundled Maven 3.9.x, which also defaults to the native transport) honors the same transport during project import. Without this, the m2e import fails PKIX, the workspace never signals ready, and analysis hangs in `Prepare()`.

## Testing

- Added `TestAppendMavenInsecureArgs` asserting both `-Dmaven.resolver.transport=wagon` and `-Dmaven.wagon.http.ssl.insecure=true` are present.
- Verified end-to-end against a Maven mirror served over a self-signed TLS certificate: with the stock provider image, dependency resolution fails PKIX and the Dependencies report is empty; with a provider image built from this branch (only variable changed), dependencies resolve, JDTLS imports the project, and the expected insights are produced with zero PKIX errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Maven connectivity for repositories using self-signed certificates or private certificate authorities when insecure Maven access is enabled.
  - Applied the required transport and TLS settings consistently across dependency resolution, downloads, local repository access, and Java language tooling.

- **Tests**
  - Added coverage to verify insecure Maven arguments are appended correctly while preserving existing options.


jmle (approved):
Looks good 👍

